PR #5, opened 2016-09-08
Change message-header-name face

cocreature:
Before:
mail_1

After:
header_0

tbh I don’t actually care about `message-header-name` but about `mu4e-header-key-face` which inherits from it. Setting `message-header-name` should help other mail clients but I haven’t tested any.


hlissner:
That looks much nicer. Thanks!

PR #10, opened 2016-09-16
NEW add elscreen

kwrooijen:
Added some colors for Elscreen. I named the variables `tab-unfocused-bg/fg` so that it can be reused for other tab libraries, since elscreen isn't the only one. Here's an image of how it looks:

image


hlissner:
Thanks!

PR #22, opened 2016-10-31
Ivy mode - a more visible selection color

rakanalh:
This change works for me as a more visible selection color for ivy-mode. 

Although i am not very sure about blue being the right color to use here. 

What do you think?


rakanalh:
@hlissner have you had the chance to look at this?

hlissner:
Sorry, I've only just found time to look at it. I think this is fine, so I will merge it.

rakanalh:
Thanks alot!

PR #27, opened 2016-11-27
Don't invert colors for show-paren-match

### LINKED ISSUE #25: blue cursor collides with show-paren-mode
1cfecf5ef9e94a94cf33f0b8408c8ef810fd527c enabled a blue cursor by default. Unfortunately this makes the cursor invisible if it is on the left parantheses and `show-paren-mode` is enabled. See the attached image where the cursor is on the left parenthesis.
leftparen

It looks like the face responsible is `show-paren-match`. I don’t have any clever suggestions on how to fix this apart from either changing the cursor back to white or finding a pretty color for `show-paren-match` that doesn’t collide.

`show-smartparens-mode` suffers from the same problem but its faces inherit from `show-paren-match` so this should fix both.

ppiotrowicz:
Fixes #25

### DIFF
--- a/doom-one-theme.el
+++ b/doom-one-theme.el
@@ -113,7 +113,7 @@
      `(match                  ((,c (:foreground ,green :background ,black :bold ,bold))))
      `(trailing-whitespace    ((,c (:background ,doc-comments))))
      `(vertical-border        ((,c (:foreground ,vertical-bar :background ,vertical-bar))))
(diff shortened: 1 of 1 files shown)


hlissner:
Hey! Sorry for the late reply. This is a step in the right direction, but it needs a little more (for instance, if you're using rainbow-delimiters, `rainbow-delimiters-depth-1-face`'s `:foreground` is blue, and will conflict with this).

So I'll have to reject the PR, but I'll experiment and see if I can come up with something.

PR #38, opened 2017-02-17
Support which-function-mode

cocreature:
Without this it’s sadly almost unreadable. I don’t really care which color in particular is chosen as long as it’s readable so if you have any better suggestions I’m all ears.

*before*
doom_which_func
*after*
doom_which_func_2


hlissner:
Looks fine to me. Thanks!

PR #42, opened 2017-03-12
Add terminal colors

elektronaut:
Hi, here are some face definitions that work with both term and ansi-term.


hlissner:
Looks good, thanks!

PR #53, opened 2017-05-06
Add basic support for auctex faces

justbur:
Before:
doom-one-before

After:
doom-one-after


hlissner:
Excellent work, thanks!

PR #56, opened 2017-05-09
change forground of swiper matches to black for better contrast

gilbertw1:
Change the foreground of swiper matches to black to allow for better contrast and visibility.

Before:

swiper-matches-before

After:

swiper-matches-after

After (molokai):

swiper-matches-after-molokai


hlissner:
Excellent! I've been meaning to do this. Thank you!

PR #57, opened 2017-05-12
Add doom-one-padded-modeline option

justbur:
I think this looks nicer, so I added an option. 

Screenshot:
padded-modeline


hlissner:
Excellent, thanks for both PRs!

PR #59, opened 2017-05-13
Add ledger faces

ppiotrowicz:
https://github.com/ledger/ledger-mode

Here's how it looks like with doom-one
test dat 2017-05-13 22-14-22


hlissner:
Excellent, thanks!

PR #63, opened 2017-05-17
Theme link face

justbur:
I chose blue, but cyan might be more appropriate (I prefer the blue FWIW).


hlissner:
Sounds good to me, thanks!

Issue #60, opened 2017-05-16
color scheme request


bigardone:
Hi! First of all thanks for the theme, I simply love it :)
Would it be possible requesting a color scheme based on Nova for vim? https://trevordmiller.com/projects/nova
If you don't have the time for it, I could work on a PR for adding it following this basic color list described here https://github.com/trevordmiller/nova-colors/blob/master/src/index.js
What do you think?

hlissner:
That is a lovely colorscheme! So I'd be happy to include it.

I've got others in the pipeline however, so it may take me a while to get to it. That basic color-list PR would be appreciated though, and would go a long way to speed the process up.

bigardone:
Awesome! I'll start working on that PR then :)
Cheers!

PR #64, opened 2017-05-19
Nova theme basic colors

bigardone:
This PR adds the basic color scheme for new theme based on [Nova](https://trevordmiller.com/projects/nova) as discussed in #60.

I've marked some of the lines with a **TODO set correct color** where I don't know what color to use. I've used this values as a guide for creating the file https://github.com/trevordmiller/nova-colors/blob/master/src/index.js.

This is the WIP result, which doesn't look very good... 🤔 

image

I hope it helps as a starting point for creating another awesome doom's version 🤘

### DIFF
--- a/doom-nova-theme.el
+++ b/doom-nova-theme.el
@@ -40,7 +40,7 @@ determine the exact padding."
    (highlight      cyan)
    (vertical-bar   (doom-darken dark-grey 0.2))
    (current-line   (doom-darken bg-alt 0.05))
-   (selection      (doom-lighten highlight 0.3))
+   (selection      (doom-lighten highlight 0.6))
    (builtin        blue)
    (comments       grey)
    (doc-comments   (doom-lighten grey 0.1))
@@ -95,8 +95,16 @@ determine the exact padding."
    ;; linum
    (linum :foreground grey :background bg :bold nil :distant-foreground nil)
    (doom-linum-highlight :foreground highlight :background bg :bold nil :distant-foreground nil)
-   (cursor :background highlight))
+   (cursor :background highlight)
 
+   ;; helm
+   (helm-selection :foreground dark-grey :background selection)
+   (helm-match     :foreground highlight)
+
+   ;; company
+   (company-tooltip-selection  :background selection :foreground dark-grey)
+
+   )
 
   ;; --- variables --------------------------
   ;; ()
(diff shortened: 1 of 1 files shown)


bigardone:
here's an updated screenshot after setting blue as the highlight color and removing the brighter source buffers from the config:

image

hlissner:
An excellent start, thanks! Would you like more time to tweak it or can I merge it?

bigardone:
Glad you liked it :)
I think I'm going to try to tweak a bit more during this week if you don't mind. 
I want to change the cursor color, the selection background, and some other minor stuff.
Cheers!

bigardone:
@hlissner this is how it looks like after the last commit.

image

I've been trying to change the cursor's background color, but it doesn't seem to work because I'm probably doing it wrong. Where can I check out how to achieve this?
Thanks in advance :)

hlissner:
The cursor's background should be controlled by the `cursor` face, [here's its base setting](https://github.com/hlissner/emacs-doom-theme/blob/master/doom-themes-common.el#L35). 

Also, I notice you are a spacemacs user. I see that spacemacs [applies its own faces on the various evil cursors](https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bdistributions/spacemacs-bootstrap/packages.el#L55) -- this may be why you are not seeing the changes.

bigardone:
Thanks for the tip, I keep improving it while I use it :)
In regards to the Spacemacs, yes I use it. Do you think it's a good idea mixing emacs and spacemacs stuff in the theme?

hlissner:
I welcome support for spacemacs' faces, if that's what you mean. Though, a disclaimer: I'm not a spacemacs user, so I won't be able to debug issues.

gilacost:
I like it 😋, looking forward for this to be merged. I wonder how does a yeipiyei preview looks like 🤔.

hlissner:
@bigardone Looks good! I've got a few big changes in store for this plugin coming up. I'll merge this once I'm done with that, likely over the weekend. Or do you need more time?

bigardone:
Sounds good! I think it is a good starting point as it is right now, and I can create more PR to improve it, so go ahead and merge it :)
Cheers!

UtamaruMonster:
Great looking theme.
Anyway to make Nova look good in terminal?
This is how it looks like with emacs -nw

screen shot 2017-09-08 at 13 01 31

PR #71, opened 2017-06-02
Sync lsp faces names with upstream

gicmo:
The typos (hightlight) got fixed upstream. So we need to adapt.

I was tempted to change the faces for -read and -write to the something like magit-diff-added-highlight and magit-diff-removed-highlight, since the original color for the faces were green and red, but I couldnt find a way to trigger the faces so I left them as is for now.


hlissner:
Oh! I missed that too. Thanks!

PR #74, opened 2017-06-09
Theme whitespace-{empty,space} using base3

justbur:
Before:
old-whitespace

After:
new-whitespace


hlissner:
Yikes, that looked awful. Nice work!

PR #76, opened 2017-06-12
highlight-indentation-face was misswritten.

indianpojken:


hlissner:
Ah, good catch. Thanks!

PR #80, opened 2017-07-01
[ledger-mode] Nicer active transaction background

ppiotrowicz:
Brings back dark background for transactions under the cursor. In one of the refactoring it got brighter.

test dat 2017-07-01 16-29-42


hlissner:
Nice work, thanks!

PR #84, opened 2017-07-11
add support for built in line number highlight face

gilbertw1:
This commit add supports for the built in line number highlighting support that was just merged into Emacs master.

Here's what it looks like:

line numbers

You can find more detail on the new line number faces here: http://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS#n436


hlissner:
Thanks for the help, but I've reworked the inheritence tree and, with it, added support for display-line-numbers, so I'll have to turn down this PR.

PR #87, opened 2017-07-20
Fix: ansi-color-names-vector: vector -> vconcat

bandresen:
ansi-color expects this type of vector ["color1" "color2" "color3" ...]

(vector sequence) returns [(seq uen ce)] while
(vconcat (sequence)) returns [seq uen ce]


hlissner:
Good catch, thanks!

PR #90, opened 2017-07-24
Add default background and foreground for term-mode

IvanMalison:


hlissner:
@IvanMalison I have no problem merging this, but do you have a screenshot or something to explain why this is necessary? I don't know where this face is used.

IvanMalison:
http://imgur.com/a/82BFF without the face below, with the face above

hlissner:
Yikes, that's awful. Consider it merged! Thanks!

PR #94, opened 2017-07-30
Change vc-modified to orange

spatz:
Atom uses yellow which is too similar to green, but orange is distinct and looks better than gray.


hlissner:
I agree, orange looks much better! Thanks!

PR #95, opened 2017-07-30
doom-one: make vertical-bar a bit brighter

spatz:


hlissner:
I assume this was suggested to increase vertical-bar's contrast. I've gone the opposite route and made vertical-bar a tad darker, because Atom One Dark's borders are dark and this suggestions makes them too light.

If you prefer the brighter vertical-bar, I'd recommend using this (after you've loaded `doom-one`):

```emacs-lisp
(doom-theme-set-faces
  'doom-one
  (vertical-border :background base4 :foreground base4))
```

PR #97, opened 2017-08-03
Require common in top level to avoid warnings

spatz:
Using `doom-themes-set-faces` causes this warning:

    Symbol’s function definition is void: doom-themes--build-face

Moving the require to top level removes this warning.


hlissner:
Good catch. Thanks!

PR #101, opened 2017-08-14
Add defcustom doom-one-light-with-solaire-mode-enabled

ztlevi:
For now, I just added the defcustom for one-light. Not sure if we should do that for all themes.😅

### DIFF
--- a/themes/doom-one-light-theme.el
+++ b/themes/doom-one-light-theme.el
@@ -85,7 +85,7 @@ determine the exact padding."
    (error          red)
    (warning        yellow)
    (success        green)
-   (vc-modified    base4)
+   (vc-modified    orange)
    (vc-added       green)
    (vc-deleted     red)
 
--- a/themes/doom-one-light-theme.el
+++ b/themes/doom-one-light-theme.el
@@ -6,17 +6,22 @@
   "Options for doom-themes"
   :group 'doom-themes)
 
-(defcustom doom-one-light-brighter-modeline nil
+(defcustom doom-one-light-with-solaire-mode-enabled t
+  "Default t. If you do not use solaire-mode, set it to nil"
+  :group 'doom-one-light-theme
+  :type 'boolean)
+
+(defcustom doom-one-light-darker-modeline nil
   "If non-nil, more vivid colors will be used to style the mode-line."
   :group 'doom-one-light-theme
   :type 'boolean)
 
-(defcustom doom-one-light-brighter-comments nil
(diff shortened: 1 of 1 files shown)


ztlevi:
😅Are we going to do this? Use a defcustom for non-solaire-mode users, like me. I am using Spacemacs and I expect this custom feature could be put in the repo so that we don't need to modify the faces in our config file.

I could drop this if u guys think it's not necessary. Or a better way doing this?

hlissner:
Sorry for the late response, I've been both sick and busy!

I'll have to turn down this PR:

1. I think the solaire-mode option(s) over-complicates things. Solaire-mode has the `solaire-mode-swap-bg` function for exactly this problem (i.e. the background and brighter faces are the wrong way around), and I'd like to avoid any more coupling than is absolutely necessary.

2. `s/brighter/darker`: I don't think this is necessary. Perhaps "vivid" would have been a better choice, but not only do I prefer consistency with the variables of other themes (where possible), but "brighter" isn't necessarily wrong here. In all cases, we're replacing the dim grey default with a color that's typically easier to see (in doom-one-light's case, cyan). I'd consider that "brighter" than dim grey, in the sense that it's more vivid.

3. Changing vc-modified to orange: I agree with this change. in fact, I'll incorporate it into doom-one-light closer to the weekend, unless you'd like to set up another PR for it. Whichever happens first.

-----

Thanks for the help though!

ztlevi:
Alright, but is there a way for us non-solaire-mode users?

hlissner:
It should work out of the box for non-solaire-mode users. That it isn't is a bug, which this latest commit should resolve.

I merged some of your new colors, changed vc-modified, and tweaked doom-one-light to better suit both solaire-mode and non-solaire-mode use alike.

To answer your question in more detail: non-solaire-mode users will see the native faces (e.g. `mode-line`, `hl-line` and `default`), while solaire-mode users will see the remapped faces (e.g. `solaire-mode-line-face`, `solaire-hl-line-face`, and `solaire-default-face`). As long as each group of faces are configured properly, you shouldn't need to do anything more.

Still, this is a WIP. Thanks for helping me spot these issues; let me know if you find more!

ztlevi:
Oh, thanks a lot!!!

ztlevi:
😅I'm sorry but I don't think the hl-line face is properly set. I'm using the latest version, but the line the basically white...
hl-line

ztlevi:
Thanks!!! Nice done.🤡

PR #116, opened 2017-11-11
fix the linum-relative-current-face for peacock-theme.

ichirika:
this is still that underline problem.lol
(btw, expecting tron-legacy and badger


hlissner:
Could you add `:underline nil` to `line-number` and `line-number-current-line` instead of `linum-relative-current-face`? `linum-relative-current-face` inherits from them in [doom-themes-common.el](https://github.com/hlissner/emacs-doom-themes/blob/master/doom-themes-common.el#L444), and I don't want to break this inheritance.

ichirika:
yeah, ok~

PR #117, opened 2017-11-13
Face tweaks for markdown and org block, helm, web-mode, wgrep

ztlevi:
Base3 is too dark for markdown, so I put base1 here. While for org block, I think it should be consistent with markdown block.

This is how it looks in doom emacs with `solaire-mode`.

jietu20171113-181745 2x

I tweaked some faces with black background as default. It looks better now.

Thanks!


hlissner:
Looks good! Thanks, I appreciate the help!

PR #112, opened 2017-11-01
Peacock Theme

teesloane:
Wanted to try my hand at making a theme. This PR is very much a wip, but I thought I'd see if I could get some feedback if I'm in the right direction.

# Process Notes

- I duplicated doom-one-theme as you suggested in #82 
- Opened up the peacock emacs theme [here](https://github.com/daylerees/colour-schemes/blob/master/emacs/peacock-theme.el) and looked through the colours in `rainbow-mode`
- I didn't expect this, but I was able to just load the theme and then continually eval the buffer and see the changes. I guess that's emacs goodness. Pretty cool.
- I arbitrarily chose some colours (`bg-alt`, for example). 
- I didn't know what to do for the the "16" coloumn of colours, so you'll see a couple `peacock todo 16` comments kicking around...
- I couldn't get operators to be red... 🤔 (see screenshot in the top right pane)
- I mucked about in the code under the `;; custom categories` a bit but didn't really know what was going on.
- I'm imagine there is a bit of irrelevant code from doom-one that can be removed but ¯\_(ツ)_/¯

Anyways, any input / feedback would be cool! I'm very new to writing elisp / writing lisps in general, so any pointers in that regard would be great too.

# screen shot

image

PR #123, opened 2017-11-16
fix the issue #112 -> Problem with company popup on peacock theme

ichirika:


hlissner:
Sorry for taking so long to get to this! Thanks for the fix!

PR #126, opened 2017-11-29
Add ediff faces and org quote background

ztlevi:
Just a slight modification. I use Spacemacs and its built-in theme has differences between `ediff-current-diff-*` faces. So I modified them in doom-one-light theme.

Here are what these faces look like.
image


hlissner:
Sorry for taking so long to get to this! This looks good, thanks!

PR #133, opened 2018-01-23
Change vertical-bar so that window-divider stands out.

legendre6891:
`vertical-bar` was set to a color too close to the background, making it essentially invisible in most dark themes. It seems like `(doom-lighten bg 0.1)` is a good choice for dark themes, and correspondingly `(doom-darken bg 0.1)` for light themes.

### DIFF
--- a/themes/doom-molokai-theme.el
+++ b/themes/doom-molokai-theme.el
@@ -51,7 +51,7 @@ determine the exact padding."
 
    ;; face categories
    (highlight      orange)
-   (vertical-bar   base2)
+   (vertical-bar   (doom-lighten bg 0.1))
    (selection      base0)
    (builtin        orange)
    (comments       (if doom-molokai-brighter-comments violet base5))
--- a/themes/doom-nova-theme.el
+++ b/themes/doom-nova-theme.el
@@ -50,7 +50,7 @@ determine the exact padding."
 
    ;; face categories
    (highlight      cyan)
-   (vertical-bar   (doom-lighten bg-alt 0.1))
+   (vertical-bar   (doom-lighten bg 0.1))
    (selection      (doom-lighten highlight 0.6))
    (builtin        blue)
    (comments       grey)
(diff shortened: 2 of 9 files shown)


hlissner:
This looks great. I'll give it a more thorough test run over the weekend before merging it. Thanks for your help!

legendre6891:
Great! Wondering if you had a chance to test the changes -- looking forward to seeing this upstream :-).

hlissner:
Sorry, forgot to merge it! Thanks again for the help!

hlissner:
I ended up tweaking a few of the vertical-bars back to dark colors because the light-dark contrast was too strong on my displays. As a compromise, I'm now using colors that are darker than they originally were, hoping to improve the contrast.

What do you think? Do you still think they are too dark?

legendre6891:
Thanks! I think they're better for sure -- though ultimately I guess the contrast is one's **_personal_** preference. For my tastes, I think the dark themes `one`, `vibrant`, `spacegrey`, and `tomorrow-night` still have too low of a contrast. However, for all but `vibrant`, what we have now should be acceptable for most people.

However, I still think that the contrast in `vibrant` is too low for many (I'm guessing). 

In general, perhaps this should be a configuration option. Namely, there should be `defcustom` (separate for each theme) `doom-*theme-name*-divider-contrast` variable, so that the `vertical-border` is set to

````
(doom-invert doom-*theme-name*-divider-contrast *default-color-we-have-now*)
````
where `doom-invert` evaluates to `doom-darken` for light themes and `doom-lighten` for dark themes.

Thoughts? I'm happy to work on this, btw.

**Comment by hlissner** [comment]
Sorry for the late reply. I've thought on this for a while. Would you mind simply using this?

```elisp
(set-face-foreground 'vertical-border (doom-lighten 'bg 0.1))
```

I hesitate to introduce `doom-*theme-name*-divider-contrast` variables; I'm not sure the feature warrants the added complexity, where a one-liner like the above will do.

PR #141, opened 2018-02-20
Add: Nord, City Lights, Darcula; Better Solarized Light

fuxialexander:
- Nord:
image

- City Lights:
image

- Darcula:
image

- Solarized Light:
image

### DIFF
--- a/themes/doom-city-lights-theme.el
+++ b/themes/doom-city-lights-theme.el
@@ -1,0 +1,161 @@
+;;; doom-city-lights-theme.el --- inspired by Atom One Dark
+(require 'doom-themes)
+
+;;
+(defgroup doom-city-lights-theme nil
+  "Options for doom-themes"
+  :group 'doom-themes)
+
+(defcustom doom-city-lights-brighter-modeline nil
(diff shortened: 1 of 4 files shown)


hlissner:
These look great, but I was bothered by how much doom-themes-common has been changed, which had a drastic effect on the other themes. 

I've cherrypicked bits and pieces into the common rules (you'll find a summary in ff0965b).

Please try out the themes and let me know if anything looks amiss.

fuxialexander:
@hlissner Ah I guess I did note correctly limit the changes to only those `theme` files when merging from my local worktree 😂I did intended to only commit the change on theme files only (if you check the "files changed", there should only by those files got involved.) Maybe I should use a file-based diff rather than directly merge from my worktree. (Or maybe I can limit the merge to some specific files, and compress commit history into a single commit? Any idea?) Sorry for my unfamiliarity with Magit (git) 😅
Anyway, I would also like to thank you for provide such an excellent theme base (which is really easy to customize) and the awesome Doom-config! My emacs is really like day vs night than one year before!

PR #145, opened 2018-03-08
Fixed typo and made dracula theme more close to original.

adamkruszewski:
Fixed filename typo for dracula theme and changed its org-mode faces to match original
more closely (only in org-mode).


hlissner:
Thanks for this! Here I was thinking darcula was an intentional, haha.

Anyhow, I will be doing some work on this soon, mostly to conform it to org-level-* conventions across all doom themes. I'll add an option to undo this, but keep it in mind if you one day update and don't like what you see!

Issue #146, opened 2018-03-15
dark-cyan is not defined in doom-citylights and doom-nord


DVN196:
Love the new themes but since dark-cyan is not defined in these two theme, when I toggle `brighter-comments` variable, I can't load the theme.

PR #148, opened 2018-03-18
Fix fg/bg and modeline-bright

fuxialexander:
This PR fixes the fg/bg without solaire-mode and modeline color when modeline-brighter is t in the following themes:
- Solarized light
- Citylights
- Nord
- Dracula


fuxialexander:
#146 Fixed in recent commit

fuxialexander:
I think it's good to go

fuxialexander:
Ahh.... just clicked the wrong button....

PR #151, opened 2018-03-22
Adjust City Lights and Nord color to comply with specification

fuxialexander:
1. Added `doom-nord-region-highlight` variable to specify the region highlight style: "forest", "snowstorm" and default
2. Adjust modeline-bright so that they looks more natural
3. Adjust specific faces (City Lights `org-block`)


hlissner:
This looks great, thanks!

PR #152, opened 2018-03-23
Fix: Byte-compilation error of nord theme

fuxialexander:
Previously the introduced `  nord-region-fg/bg` colors break the byte-compilation. This PR removed those color specification and merge them with the `region` color.

### DIFF
--- a/themes/doom-nord-theme.el
+++ b/themes/doom-nord-theme.el
@@ -82,15 +82,6 @@ determine the exact padding."
    (variables      (doom-lighten magenta 0.5))
    (numbers        magenta)
    (region         base4)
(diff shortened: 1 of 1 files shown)


hlissner:
I've pushed a slightly different approach to this fix, with one important change: `doom-nord-region-highlight` now takes symbols instead of strings (a convention I'd like to stick to with multiple and finite choice variables).

I've also refactored `doom-nord-region-highlight` to reduce redundancy. The `&override` keyword makes that rule inherit the rule from `doom-themes-common.el`.

Could you try this and let me know if that gives you byte-compilation issues?

PR #154, opened 2018-03-25
Making tags color more consistent.

adamkruszewski:
A very minor change but it makes overall effect much better.


hlissner:
I'm not sure I agree with this change. There are many elements in dracula's org-mode that are orange, like links, counts and TODO's.

image

adamkruszewski:
You are wight, light orange looks much better.

hlissner:
Thanks!

PR #155, opened 2018-03-25
Darker comments possibility for Solarized Light

macaroot:
In the Solarized Light theme, base5 is used for comments, and it is defined extremely light to my eyes (Also atypically light compared to standard Solarized Light, I think). I would like to have the option to have the comments use base6 instead. I would also want to use base6 for doc-comments and modeline-fg-alt, but the most crucial is comments, so that's what I'm addressing here.

This uses a bit cruddy short circuit with doom-solarized-light-brighter-comments, but I didn't want to merge the defcustoms or do anything more complicated before I knew if this was any way desirable for anyone else.

This is what comments look like with the original base5:
normal

This is what they would look like with base6:
darker

### DIFF
--- a/themes/doom-solarized-light-theme.el
+++ b/themes/doom-solarized-light-theme.el
@@ -28,10 +28,11 @@ be displayed in slightly darker colors."
   :group 'doom-solarized-light-theme
   :type 'boolean)
 
-(defcustom doom-solarized-light-darker-doc-comments nil
-  "If non-nil, doc-comments will be displayed in slightly darker colors."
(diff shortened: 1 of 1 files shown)


hlissner:
I agree. The comment colors have poor contrast. I've gone in and done a thorough revision of solarized-light's colorscheme, including the comments. Please update and give it a try.

macaroot:
@hlissner As someone who has used Solarized Light for everything from terminal and window manager up, I feel a bit more at home now.

Btw doom-emacs has become my window manager (with exwm), so I'm thankful.

PR #168, opened 2018-04-23
Add cperl faces

sonologico:
I kept the bold for arrays and bold/italic for hashes, but removed the jarring backgrounds.

Old:
old

New:
new


hlissner:
Good catch, thanks!

PR #172, opened 2018-04-30
Add: update posframe faces

ztlevi:
The original background is so dark that I cannot even recognize the line selection.

I make the background brighter like this.

image


hlissner:
Thanks!

PR #173, opened 2018-05-02
Add doom-themes-padded-modeline custom var

justbur:
Serves as the default value for the padded-modeline variable for themes that
have one


hlissner:
It does make more sense to have a single variable for this. I'll make the theme-specific ones obsolete and remove them after some time. Thanks!

PR #188, opened 2018-05-31
Fix spacegrey color definition.

jsoo1:
I was getting a lot of messages about not being able to load the color.  It was just missing a hashmark in front of the hex code 😄 .  These themes are great btw!


hlissner:
Ah! A careless mistake. Thanks for catching it!

PR #199, opened 2018-06-15
A few updates for lsp

ztlevi:
Just make lsp highlight face be consistent. Since `ivy-posframe-cursor` now inherit from cursor face, we can remove it now.

Thanks 😃


hlissner:
Good work. Thanks again!

PR #201, opened 2018-06-25
Add: org-indent face

alphapapa:
This fontifies org-indent to match the `org-default` face so that indentation characters match the color of the background.  This is important in case the org-indent character is not a space (e.g. when using variable-pitch fonts in Org buffers, using a * for indentation is better than a space, because it aligns entry content with headlines, while spaces result in content that is not indented enough).

### DIFF
--- a/doom-themes-common.el
+++ b/doom-themes-common.el
@@ -1085,6 +1085,7 @@
     (org-formula                  :foreground cyan)
     (org-headline-done            :foreground base5)
     (org-hide                     :foreground bg)
+    (org-indent                   :foreground bg)
 
     (org-level-1 :foreground blue     :background base3 :weight 'ultra-bold :height 1.25)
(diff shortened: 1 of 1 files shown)


hlissner:
Has the org-indent face changed in recent versions of org-mode? Mine (version 9.1.13) is:

```elisp
;; in org-indent.el (line 107)
(defface org-indent '((t (:inherit org-hide)))
  "Face for outline indentation.
The default is to make it look like whitespace.  But you may find it
useful to make it ever so slightly different."
  :group 'org-faces)
```

Meaning it inherits from `org-hide`. That is why I omitted a definition for it. Perhaps this indicates a different issue?

alphapapa:
Yes, I think the issue is that I had customized the face a long time ago and forgotten about it, and now the customization is unnecessary.  Thank you, and apologies for the error.

PR #206, opened 2018-07-23
make org-level inherit default in order to have proper font

ztlevi:
`Org-level`'s font looks different for other languages. This one looks more apparent.

image

And applying the default inheritance fix the issue.


ztlevi:
Yes. Thanks. It turns out it's due to the `ultra-bold` attribute in `doom-one-light`. I replace them with just `bold`.

hlissner:
As of 4c2ee8d and ff7f595 this should be fixed. Sorry for the wait!

ztlevi:
Gotcha. Thank you!

Issue #226, opened 2018-10-10
Is this a mistake about gnus-groupnews-1-empty?


nasyxx:
Hi, I meet a bug that when I generate org htmlize css ( <kbd>M-x org-html-htmlize-generate-css</kbd> ):

```emacs-lisp
Debugger entered--Lisp error: (error "Invalid face" gnus-groupnews-1-empty)
  internal-get-lisp-face-attribute(gnus-groupnews-1-empty :inherit nil)
  face-attribute(gnus-groupnews-1-empty :inherit)
  htmlize-face-size(gnus-group-news-3-empty)
  htmlize-face-to-fstruct-1(gnus-group-news-3-empty)
  #f(compiled-function (face) #<bytecode 0x6aaa5c11>)(gnus-group-news-3-empty)
  mapcar(#f(compiled-function (face) #<bytecode 0x6aaa5c11>) (gnus-group-news-3-empty))
  htmlize-face-to-fstruct(gnus-group-news-3-empty)
  htmlize-make-face-map((default bold italic underline fixed-pitch fixed-pitch-serif variable-pitch shadow link link-visited highlight region secondary-selection trailing-whitespace line-number escape-glyph homoglyph nobreak-space nobreak-hyphen mode-line mode-line-inactive mode-line-highlight mode-line-emphasis mode-line-buffer-id header-line header-line-highlight vertical-border window-divider window-divider-first-pixel window-divider-last-pixel internal-border minibuffer-prompt fringe scroll-bar border cursor mouse tool-bar menu help-argument-name glyphless-char error warning success read-multiple-choice-face tty-menu-enabled-face tty-menu-disabled-face tty-menu-selected-face show-paren-match show-paren-match-expression ...))
  htmlize-buffer-1()
  htmlize-region(1 817)
  org-html-htmlize-generate-css()
  funcall-interactively(org-html-htmlize-generate-css)
  call-interactively(org-html-htmlize-generate-css record nil)
  command-execute(org-html-htmlize-generate-css record)
  helm-M-x(nil #("org-html-htmlize-generate-css" 0 29 (match-part "org-html-htmlize-generate-css")))
  funcall-interactively(helm-M-x nil #("org-html-htmlize-generate-css" 0 29 (match-part "org-html-htmlize-generate-css")))
  call-interactively(helm-M-x nil nil)
  command-execute(helm-M-x)
```

[...]

PR #227, opened 2018-10-10
Replace gnus-groupnews-1-empty to gnus-group-news-1-empty

nasyxx:
As I said in #226 , when I change it, it becomes working well.


hlissner:
Thanks for your help!

PR #228, opened 2018-10-10
Change bg-alt for tomorrow-night to match atom

benwtks:
Very minor change but one that I think would look better

### DIFF
--- a/themes/doom-tomorrow-night-theme.el
+++ b/themes/doom-tomorrow-night-theme.el
@@ -16,7 +16,7 @@ determine the exact padding."
 
   ;; name        gui       256       16
   ((bg         '("#1d1f21" nil       nil          ))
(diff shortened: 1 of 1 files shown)


hlissner:
This breaks solaire-mode support. Did you mean to change `bg`, instead of `bg-alt`?

benwtks:
#161719 is supposed to be for the side bar, magit, helm, etc. I had to use bg swap to get it the right way around so maybe it's in the wrong place but II also have to do that with other themes like dracula, city lights and one dark, they all have the solaire thing the wrong way around without the swap. I looked at the other themes and saw the darker one is usually bg-alt so that's what I did

hlissner:
Ah, I understand. Yes, bg/bg-alt are the other way around, as not to impose the darker background on non-solaire-mode users as a default. In any case, thanks for the PR!

Issue #233, opened 2018-10-15
molokai theme: some color hard to see because of low contract


laishulu:
molokai theme: some color hard to see because of low contract 
1. forground of git-gutter sign (`cyan` looks good as I test)
<img width="251" alt="snipaste_2018-10-15_23-29-41" src="https://user-images.githubusercontent.com/1224321/46961098-91a4b400-d0d2-11e8-9678-8eba2174c9df.png">
2. backgournd of mini-buffer and company selection (brightblack` looks good as I test)
<img width="313" alt="snipaste_2018-10-15_23-30-13" src="https://user-images.githubusercontent.com/1224321/46961337-227b8f80-d0d3-11e8-95df-863d1a7ccf5b.png">
<img width="327" alt="snipaste_2018-10-15_23-37-25" src="https://user-images.githubusercontent.com/1224321/46961427-50f96a80-d0d3-11e8-8cec-5d94f9c4cfe0.png">

laishulu:
I don't know how to modify theme directly, or I will send you a pull request to offload your burden.

laishulu:
@hlissner please merge https://github.com/hlissner/emacs-doom-themes/pull/236

PR #236, opened 2018-10-16
fix: #233

laishulu:


hlissner:
Ah, I understand now why base4 is too difficult to read. My bad, I had forgotten about terminal users. In GUI Emacs I use fringe bitmaps, e.g.

image

In which case, cyan becomes too loud. It may be better to give `git-gutter:modified` a different color from `git-gutter-fr:modified`, instead of changing `vc-modified` overall. I'll look into that.

For the time being, could you undo that one change?

laishulu:
@hlissner modified following your instruction

laishulu:
please merge if no problem.

laishulu:
please merge.......

hlissner:
Sorry for taking so long, I was away on a trip. Thanks for the PR!

PR #238, opened 2018-11-01
Add background for ivy-current-match

kevinjfoley:
Can't see what line is currently selected when using swiper as is


ema2159:
I think this isn't a problem now, I've fixed this theme. 
Just an advice, try to not hard code colors when submitting PRs, it's considered a bad practice. Try to use the already existing colors from the palette.

PR #249, opened 2018-12-13
Add vterm support based on term

justbur:
See https://github.com/akermu/emacs-libvterm


hlissner:
Thanks for your help!

PR #254, opened 2019-01-01
Add TeX-error-description-* faces

d125q:
These faces are used when invoking `` C-c ` `` to display the list of errors from a TeX-mode buffer.  I modeled them after the compilation-mode faces, but do not want to inherit since `compile.el` will not necessarily be loaded.


hlissner:
Good work. Thanks for your help!

PR #255, opened 2019-01-10
[RFC] IR Black

legendre6891:
I took the colors from the well-known `ir_black` palette. Unfortunately, this was a while ago, so the structure may need some cleanup. Let me know what needs to be done.

### DIFF
--- a/themes/doom-ir-black-theme.el
+++ b/themes/doom-ir-black-theme.el
@@ -1,15 +1,10 @@
-;;; doom-ir-black-theme.el --- port of the original ir_black theme
+;;; doom-ir-black-theme.el --- port of the original ir_black theme -*- lexical-bindings: t; no-byte-compile: t; -*-
 (require 'doom-themes)
 
 ;;
 (defgroup doom-ir-black-theme nil
   "Options for doom-themes"
   :group 'doom-themes)
-
-(defcustom doom-ir-black-smooth-bg nil
-  "If non-nil, comments will be highlighted in more vivid colors."
-  :group 'doom-ir-black-theme
-  :type 'boolean)
 
 (defcustom doom-ir-black-brighter-comments nil
   "If non-nil, comments will be highlighted in more vivid colors."
@@ -30,12 +25,11 @@ determine the exact padding."
 
 ;;
 (def-doom-theme doom-ir-black
-  "Port of the original IR black theme"
+  "A dark theme inspired by Atom One Dark"
 
   ;; name        default   256       16
-  ((bg         (if doom-ir-black-smooth-bg '("#232323" nil nil)
-		  '("#000000" nil nil)))
-   (bg-alt     '("#121212" nil       nil            ))
+  ((bg         '("#000000" "black"   "black"        ))
+   (bg-alt     '("#121212" "black"   "black"        ))
    (base0      '("#1B2229" "black"   "black"        ))
    (base1      '("#1c1f24" "#1e1e1e" "brightblack"  ))
    (base2      '("#202328" "#2e2e2e" "brightblack"  ))
@@ -161,4 +155,3 @@ determine the exact padding."
   )
 
 ;;; doom-ir-black-theme.el ends here
-
--- a/themes/doom-ir-black-theme.el
+++ b/themes/doom-ir-black-theme.el
@@ -6,11 +6,6 @@
   "Options for doom-themes"
   :group 'doom-themes)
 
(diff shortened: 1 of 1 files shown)


ema2159:
Damn. did't noticed this was a theme PR. I'll try to review it soon.

legendre6891:
This is based off of IR_Black, which you can find here https://github.com/twerth/ir_black

Cheers

ericsoderstrom:
Any chance we could get this merged? Love this theme

legendre6891:
I made this theme and the PR a long time ago, but would also like to see it merged. Happy to bring it up to date to the current iteration if that would help push it through (maybe @hlissner could chime in with suggestions on how to proceed)?

ema2159:
I'm sorry! I completely forgot about this. I have a question though. Where were you grabbing the colors from for this theme? I've been looking at other implementations and though the palette looks similar, there are discrepancies like the background for example.
Yours:
image

Other Emacs port (https://github.com/jmdeldin/ir-black-theme.el)
image

legendre6891:
No worries @ema2159 , thanks for taking a look. The colors are taken from the corresponding vim theme — you can see there is a “ir dark” version whose background is slightly less saturated than pure black, as per the screenshot.

I made / adapted this theme for Emacs almost 3 years ago, so my memory is a bit hazy: I think I also did some manual tuning of the colors to make it look a bit nicer to my tastes, owing to the different rendering of the same color between emacs GUI and whatever terminal I was using vim on at the time. 

Cheers

ema2159:
Can I grab the theme and try to transform to be as close to the original IR Black? It would be great since doom-themes has no black background theme as mentioned in #590, which would improve the package's accessibility.

legendre6891:
Yes, @ema2159  please go ahead. I think what would be really great is if we could have both my slightly more subtle background and a “pure” black background as you mentioned. 

The reason is that depending on the monitor type, a non-fully saturated background may actually be more readable than pure black — I know this since my laptop is ~8 years old and its monitor is old enough that pure black doesn’t work well. This is probably the intent behind the issue you linked.

as for modifying, you may find it helpful to start with 

https://github.com/twerth/ir_black/blob/master/colors/ir_dark.vim

https://github.com/twerth/ir_black/blob/master/colors/ir_black.vim

which are the original creators themes for vim. My background colors come from ir_dark I believe.

cheers

ema2159:
Got it! I will try

ema2159:
I created a customization variable that allows the user to choose between a completely dark background (default), and the slightly lighter background. The theme is ready to be merged, but there's one last thing to do. Could you add yourself to the credits section in the README and in the doom-themes.el file?

themes/doom-ir-black-theme.el
```diff
@@ -0,0 +1,164 @@
+;;; doom-ir-black-theme.el --- port of the original ir_black theme
```
hlissner, line 1: Please add ` -*- lexical-bindings: t; no-byte-compile: t; -*-`

This will fix the `Bytecode overflow` errors in the tests.

themes/doom-ir-black-theme.el
```diff
@@ -0,0 +1,164 @@
...
+;;
+(defgroup doom-ir-black-theme nil
+  "Options for doom-themes"
+  :group 'doom-themes)
+
+(defcustom doom-ir-black-smooth-bg nil
+  "If non-nil, comments will be highlighted in more vivid colors."
+  :group 'doom-ir-black-theme
+  :type 'boolean)
+
```
hlissner, line 13: Would you mind removing this? I ask because it's relatively trivial for the end users to customize with:

```elisp
(custom-set-faces
  `(default ((t (:background "#000000")))))
```

I'm also resistant to adding variables for every minor change.

hlissner (changes requested):

themes/doom-ir-black-theme.el
ema2159: Sorry. I was the one who added it. I added it on his request but I'll remove it then!

hlissner: No worries! I've only started (quietly) cracking down on our themes a couple days ago.

ema2159: I noticed! My email won't stop notifying me haha.

ema2159:
@legendre6891 Thank you very much for your contribution man! It took a (maybe excessively) long time, but it is finally there.

legendre6891:
Great, thanks a lot everyone!!

PR #267, opened 2019-03-11
one-light: don't set solaire-hl-line-face

luigy:
instead fallback to the one set from doom-themes-common

otherwise in buffers like treemacs it results in the same background color as hl-line face


hlissner:
Thanks for your help!

PR #269, opened 2019-03-19
Set treemacs closed folder icon to not slanted

jhacksworth:
`treemacs-icon-closed-png` inherits attributes from `font-lock-doc-face`, which may be italicized in some themes, e.g. `doom-one-light`. This causes the closed folder icon to appear slanted in those themes.

This commit explicitly sets the face to `:slant normal`, so that closed folders are never italicized/slanted.

For example, doom-one-light treemacs _before_:

before

And _after_ (note the folders are no longer slanted):

after


hlissner:
Sorry for the late response and thanks for the help!

jhacksworth:
Cheers! 🍻

PR #275, opened 2019-04-22
Add faces and vars for objed

flatwhatson:
Adds support for https://github.com/clemera/objed


hlissner:
Thanks again for your help!

PR #282, opened 2019-05-09
Fix #281 : diredfl is ugly

### LINKED ISSUE #281: diredfl is ugly
This is by no means an important problem, and no excessive amount of time should be spent trying to solve it. In fact, this issue is merely an introduction to a pull request that will be linked in the comments.

There are ugly rectangles around the letters when in diredfl-mode with a doom theme loaded. This is for all the themes, but it is most visible for dark themes :

2019-05-09-172808_1914x1056_scrot

2019-05-09-172943_1914x1056_scrot

2019-05-09-173024_1914x1056_scrot

This comes from the fact that all the diredfl faces edited in the doom-themes-common.el use bg-alt instead of bg as background face. The bg-alt color being slightly different than bg (being the color of the rest of the buffer), those rectangles appear.

This state of affairs has been introduced by commit ff0965b, more than a year ago. I found it funny at first that I was the first to report the problem, but then :
- diredfl is not used in the doom emacs distro, which means the number of people using both can be small
- It is only really visible on dark themes 
- Maybe people don't care
- Maybe bg-alt was used instead of bg for a very good reason

Anyway, my solution to the problem is to replace bg-alt with bg in the diredfl faces, which solves the problem. I made that change for me, but I'm pushing it upstream, since it might be useful to others (it would also be nice to have this change in master, since I could stop worrying about it).

Anyway, the PR follows. Keep me posted.

Aymeric Agon-Rambosson

aagon:
Fix for #281.


hlissner:
Sorry for the late response and thanks for looking into this for me. It appears diredfl faces came in from a PR, by a user who uses [solaire-mode](https://github.com/hlissner/emacs-solaire-mode), which uses `bg-alt` for non-file-visiting buffers (like dired).

In any case, I don't think these faces need a `:background` attribute at all. Could you try removing them?

aagon:
After a quick test, it seems safe to remove all the ":background bg" attributes, since they're actually redundant with the default face (defined on line 15 of the file doom-themes-common.el). However, I would keep the others that actually change the face for deletion-marked (reddish) or marked (yellowish) entries.

I change the PR branch accordingly.

hlissner:
Looks good to me. Thanks again!

PR #284, opened 2019-05-10
Fix #283 : Fixing and refactoring molokai

### LINKED ISSUE #283: The new doom-modeline release has broken molokai
The latest release 2.1.0 of doom-modeline has broken the molokai theme, **but only that one !**

More specifically, parts of the modeline that were previously polychromatic are now monochromatic. The bug can be reproduced by simply loading doom-molokai on startup (no other previous theme loaded !) while having the latest (2.1.0) version of doom-modeline available.

The bug stems from the fact that the modeline foreground was hardcoded to a colour (base8), instead of nil. The molokai theme, as far as I could check, is the only one where this is the case.

The molokai theme also seems not to follow some structure conventions that ~~all other themes~~ the other themes I bothered checking follow.

The mono-colouring only appeared after Mr. Zhang's new implementation of the doom-modeline at 4795a3b (it may have something to do with the fact that he generalised the use of :inherit to both active and inactive modelines)

Anyway, having the modeline behaving as it did before merely requires to replace base8 with nil in the mode-line :foreground statement.

I implemented those changes, and I used the opportunity to make the structure of the molokai theme more coherent with that of the others, without changing its behaviour (making aesthetic choices is not my place).

A pull request will follow.

Aymeric Agon-Rambosson

aagon:
Fix for #283.

### DIFF
--- a/themes/doom-molokai-theme.el
+++ b/themes/doom-molokai-theme.el
@@ -6,8 +6,19 @@
   "Options for doom-molokai."
   :group 'doom-themes)
 
+(defcustom doom-molokai-brighter-modeline nil
+  "If non-nil, more vivid colors will be used to style the mode-line."
+  :group 'doom-molokai-theme
+  :type 'boolean)
+
 (defcustom doom-molokai-brighter-comments nil
   "If non-nil, comments will be highlighted in more vivid colors."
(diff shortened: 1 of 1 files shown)


hlissner:
Nice work! Thanks for looking into this for me.

PR #285, opened 2019-05-19
Register treemacs-tags-face to also get variable-pitched

lepisma:
Not sure if this was intentional, but tags in non-variable pitch face look out of place when everything else is variable-pitched.


hlissner:

PR #298, opened 2019-07-08
Fix issue with incorrect bg colors in terminal emacs

yyadavalli:


hlissner:
They were set to `nil` so that Emacs' background would be transparent in the terminal, thus "inheriting" the background color of the terminal. If the background color is incorrect, shouldn't the terminal's background be corrected instead?

yyadavalli:
Sorry, I misunderstood what happens with `nil`. I thought it was defaulting to white. I was able to fix the colors in terminal.

Issue #301, opened 2019-07-12
Doom-tomorrow-night unfocused Centaur tab almost invisible


ianpan870102:
I really love doom-tomorrow-night, however when using its centaur theming the unfocused tabs are almost invisible (foreground way too dark). I'll attach a picture for reference. There are currently 4 tabs open, but one could see only `app.js` (unless you look very closely, you'll see `index.html`, `sw.js `etc)

<img width="1920" alt="Screenshot 2019-07-12 at 2 46 44 PM" src="https://user-images.githubusercontent.com/36686900/61108054-f0595b80-a4b3-11e9-8d97-7e990994fc60.png">

ianpan870102:
Closing this issue as I've solved this in PR #302

PR #302, opened 2019-07-14
Tomorrow-Night: change fg-alt darken to increase visibility

ianpan870102:
I have solved my problem in issue #301 by changing the value of fg-alt from being darkened by 0.6 to 0.4.

Here's a quick before-and-after: **Notice the tabs font colors**

## Before

61108054-f0595b80-a4b3-11e9-8d97-7e990994fc60

## After

Screenshot 2019-07-14 at 5 25 03 PM

### DIFF
--- a/themes/doom-tomorrow-night-theme.el
+++ b/themes/doom-tomorrow-night-theme.el
@@ -27,7 +27,7 @@ determine the exact padding."
    (base7      '("#969896" "#979797" "brightblack"))
    (base8      '("#ffffff" "#ffffff" "white"      ))
    (fg         '("#c5c8c6" "#c5c5c5" "white"))
-   (fg-alt     (doom-darken fg 0.6))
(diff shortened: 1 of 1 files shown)


ema2159:
I think the PR should be renamed to mention the changes are made in `doom-tomorrow-night` only so mr. Lissner is able to see it. 

By the way, it's important to mention that this PR matters because in all themes both `fg` and `fg-alt` faces should be visible in both `bg` and `bg-alt` . In this case, `fg-alt` isn't visible when it is over `bg-alt` so that's the error that this PR fixes.

ianpan870102:
> By the way, it's important to mention that this PR matters because in all themes both fg and fg-alt faces should be visible in both bg and bg-alt . In this case, fg-alt isn't visible when it is over bg-alt so that's the error that this PR fixes.

Thanks for mentioning that!

ema2159:
Hi Ian! I'm merging this

PR #307, opened 2019-07-20
Fix a bug with mismatching (provide) and filename

TheZeus121:
If not using MELPA for doom-themes, require didn't work properly with the org-mode extension because of mismatching filename and `(provide)`.


hlissner:

PR #309, opened 2019-07-26
Add color helper functions

casouri:
I wrote a helper function `doom-gencolor` that generates 8-bit and 4-bit colors automatically from a true color.

It works like this:
```elisp
(doom-gencolor "#ffaa33")
("#ffaa33" "#ffaf5f" "#FFFF00")
```

So I can write:
```elisp
   (red        (doom-gencolor "#e45649"))
   (orange     (doom-gencolor "#da8548"))
   (green      (doom-gencolor "#50a14f"))
   (teal       (doom-gencolor "#4db5bd"))
   (yellow     (doom-gencolor "#986801"))
   (blue       (doom-gencolor "#4078f2"))
   (blue-alt   (doom-gencolor "#a0bcf8"))
   (magenta    (doom-gencolor "#a626a4"))
   (violet     (doom-gencolor "#b751b6"))
   (cyan       (doom-gencolor "#0184bc"))
```

Instead of typing colors by hand.

### DIFF
--- a/doom-themes.el
+++ b/doom-themes.el
@@ -391,6 +391,118 @@ theme face specs. These is a simplified spec. For example:
     (add-to-list 'custom-theme-load-path
                  (or (and (file-directory-p dir) dir)
                      base))))
+;;; Auto color
+
+(defun doom-gencolor (color)
+  "Return three colors derived from COLOR: (COLOR 8-bit 4-bit)."
+  (list color (doom-closest-8-bit-color color) (doom-closest-4-bit-color color)))
+
+(defvar doom--8-bit-color-list
+  '("#000000" "#800000" "#008000" "#808000" "#000080" "#800080"
+    "#008080" "#c0c0c0" "#808080" "#ff0000" "#00ff00" "#ffff00"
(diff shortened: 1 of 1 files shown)


gagbo:
I'm not a huge fan of finding the closest color with this method (closest distance in the RGB space).

I prefer using euclidean distance in the HSV space, since having closer distances (especially in hue) feel closer to me.

I have a python script doing it : https://gist.github.com/gagbo/7943c9a71cab22b641d2904a6d59ec3a

And I started writing the elisp version of it but it's not working currently, because the comparison I use for the distance doesn't consider the fact that hue is defined modulo 360

https://gist.github.com/gagbo/565c1ea49d4230278ec5da19f912014d

hlissner:
Thanks for the PR, but I'm working on a library to help folks build these color lists while writing themes, which is a bit redundant with this PR, so I must turn it down. I'd also prefer they don't offload this work onto Emacs at runtime (all the compile-time work our themes already do is a bit much, and something I plan to fix). In any case, this work is very useful as a reference, so thank you for your help!

PR #311, opened 2019-07-27
Treemacs colorful icon theme

ema2159:
Given it's a little heavy change, I decided it was better to assign a reviewer. This change adds a `doom-themes-treemacs-color-icons` flag that when true, displays a more colorful Treemacs theme than the original, minimal atom inspired implementation by Henrik.

### DIFF
--- a/doom-themes-ext-treemacs.el
+++ b/doom-themes-ext-treemacs.el
@@ -19,6 +19,10 @@ variable-pitch face."
   :type 'integer
   :group 'doom-themes-treemacs)
 
+(defcustom doom-themes-treemacs-color-icons nil
+  "If non nil, display a lesss miimalistic Treemacs theme with colorful icons."
+  :type 'boolean
+  :group 'doom-themes-treemacs)
 
 ;;
 ;;; Library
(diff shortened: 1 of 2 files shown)


doom-themes-ext-treemacs.el
```diff
@@ -94,81 +98,310 @@ variable-pitch face."
   (treemacs-create-theme "doom"
```
flatwhatson: Perhaps it should be two separate treemacs themes?

ema2159: I don't know, what do you think? How would you do it?

ema2159: Like maybe in a separate file `doom-themes-ext-treemacs2.el` with a `(doom-themes-treemacs-config2)`?

doom-themes-ext-treemacs.el
```diff
@@ -19,6 +19,10 @@ variable-pitch face."
   :type 'integer
   :group 'doom-themes-treemacs)
 
+(defcustom doom-themes-treemacs-color-icons nil
+  "If non nil, display a lesss miimalistic Treemacs theme with colorful icons."
```
Emiller88: `"If non nil, display a less minimalistic Treemacs theme with colorful icons."`

doom-themes-ext-treemacs.el
Emiller88: I'm in favor of either, the current way makes it easy for the user to flip it on and off though, and not enable both.

doom-themes-ext-treemacs.el
ema2159: Fixed!

doom-themes-ext-treemacs.el
ema2159: Yeah, maybe it's better. The only way it could affect is for future themes, but I'm not quite sure if that makes sense (to implement more Treemacs themes)

flatwhatson: I was just thinking two calls to `treemacs-create-theme`, making separate `doom-atom` and `doom-colors` themes, and changing the last line to `(treemacs-load-theme (if doom-themes-treemacs-color-icons "doom-atom" "doom-colors")`.  They don't need to live in separate files.

Just a suggestion, I'm not strongly opposed to how it is now, just thinking about how best to organise it.

ema2159: Ok, I think what you say makes sense, the thing is that they're both the same theme but it's just the code file icons that change, but what you say it's right. Let me figure it out

flatwhatson: Otherwise if it's a variant of the same theme, I think the conditional scope should be reduce so we're not duplicating definitions on unaffected icons like "repo" and "tag".

ema2159: There's indeed some redundancy. Thanks for the heads up. The tag does change, but there's some thing that doesn't like the media icons and some other stuff. Let's make a decision, shall I separate them in two themes or leave it as a variant?

flatwhatson: Perhaps variant is better, at least it will be shorter if the duplication is resolved.

Emiller88: Would it be helpful to maybe abstract out so that anyone could make a treemacs theme?

ema2159: @Emiller88 maybe, but the effort would be too much given that I don't see another Treemacs theme implementation coming anytime soon

ema2159: Removed redundancy. I think it's ready to merge but please triple check. I'm constantly using it to test it and until now I haven't found errors.

ema2159:
> Looks good, but you've introduced tabs indentation where previously it's all spaces. Please untabify!

Sorry, didn't noticed that. Done

flatwhatson:
Thanks, looks good to me!  This just needs someone to install it and double-check that it's working.  I haven't managed to do that tonight, will do it tomorrow if it's still pending.

ema2159:
@flatwhatson don't worry! Thanks for reviewing it!

flatwhatson:
Testing it out, I'm not sure if the usability don't is quite right.  I expected the following to work:

 - set `doom-themes-treemacs-color-icons` to `t`
 - run `M-x load-theme` and load/reload a doom theme
 - see pretty new icons in treemacs

Instead, I need to do the following:

 - set `doom-themes-treemacs-color-icons` to `t`
 - run `M-x treemacs-load-theme` and select a *different* theme
 - run `M-x treemacs-load-theme` and select the doom theme
 - see pretty new icons in treemacs

@hlissner should we force a treemacs theme reload when loading a doom theme?

ema2159:
@flatwhatson the thing is that the the treemacs theme is loaded independently from the themes (because it's optional), so there's no way that the treemacs theme is loaded automatically when loading a doom theme

flatwhatson:
Looking at this again, I think the problem is that the whole `treemacs-create-theme` call needs to be re-evaluated after toggling `doom-themes-treemacs-color-icons`.  This isn't easily possible at runtime, it just runs once when this module is loaded.

The obvious solution that comes to mind is to go back to the idea of having two completely separate treemacs themes, with the `doom-themes-treemacs-color-icons` variable simply deciding which of those is loaded by default.  Then users can still switch at runtime using `treemacs-load-theme`.

ema2159:
Let's make Henrik decide! I still think this is the best option given that I don't think people will switch treemacs themes in runtime, but let's see what he thinks

flatwhatson:
OK, though I'm not sure if we really need Henrik here.  If you split into two themes, I'll merge it.

A simple use-case for changing theme at runtime is: I want to compare the two before making a final decision.  Currently that requires editing config, saving, restarting emacs... much less convenient than the couple of calls to `treemacs-load-theme` it should be.

While the current version is more compact, it's not worth sacrificing usability IMO.

seagle0128:
I implemented my own colorful themes based on `all-the-icons- already.

Refer to https://github.com/seagle0128/.emacs.d/blob/master/lisp/init-ui.el#L161. The extensions are not hard-coded.

ema2159:
@seagle0128 That's very nice and compact! The thing with our implementation is to have granularity in order to have as much control as possible with the icons given that there are certain modes or file types (i.e. Verilog) that have icon but the icon isn't associated with the file type or the mode in all the icons

ema2159:
@flatwhatson sorry I forgot to answer! I think I'll wait for Henrik since we have opposite opinions so I think he can decide what's best!

seagle0128:
@ema2159 IMO, it's better leveraging `all-the-icons-icon-alist` and `all-the-icons-mode-alist`, but it needs making big changes for some macros and functions of `treemacs`. @hlissner doesn't want to change them for now.

Of course the current changes also can make it and I glad to integrate to Centaur.

seagle0128:
Any update for this PR?

ema2159:
I'm waiting for @hlissner, he was going to be a little busy this week

ema2159:
@seagle0128 how do you use all this: https://github.com/seagle0128/.emacs.d/blob/7d14f4c7336f4901f066a5ab484372098c9b8447/lisp/init-ui.el#L139 without getting a `Keyword argument ... not one of (:icon-directory :config :extends)` error?

seagle0128:
@ema2159 I didn't use `treemacs-create-icon` for buffer icons, since the macro `treemacs-create-icon` doesn't support this case. Refer to 
 https://github.com/Alexander-Miller/treemacs/issues/504#event-2520987027. So I leverage the hash table inside of `treemacs-create-icon` to implement it.

ema2159:
@seagle0128, so you modified the Treemacs macro in order to work that way? Such a shame, we were going to use your implementation and just add the corner cases after because it was so much cleaner!!!

seagle0128:
@ema2159 Not exactly. I didn't modified the macro directly, while extracted and leveraged the inside codes to make it. You can compare my implementation and `treemacs-create-icon` macro.

ema2159:
I was missing a `(progn` jaja, it works smoothly now

seagle0128:
Perfect!

flatwhatson:
@ema2159 I've done some small cleanup and revised how the treemacs theme is configured.

Tested that the theme looks goood, and can be switched at runtime with `treemacs-load-theme`.

I think this is ready to go, so if you're happy with my changes please go ahead and squash it!

ema2159:
@flatwhatson great! I'll do it today!

ema2159:
@seagle0128 there ya go, now we have a cool colorful Treemacs theme. Thanks for your code snippet, it made the implementation smoother an cleaner

seagle0128:
@ema2159 Welcome! It looks great, and I've already updated my configurations.

PR #325, opened 2019-08-21
Add: Solarized faces for man and woman

bricewge:
2019-08-20T21:43:31+02:00
2019-08-21T03:43:29+02:00


ema2159:
There's no Man/Woman faces for all themes overall. I think it would be better to add them to the `doom-themes-base.el` file! So go ahead and add them there.

PR #326, opened 2019-08-21
Add: common faces for man and woman

bricewge:
Related with https://github.com/hlissner/emacs-doom-themes/pull/325.

Note that themes `palenight` and `moonlight` already set their own faces for `man`.

### DIFF
--- a/doom-themes-base.el
+++ b/doom-themes-base.el
@@ -1138,6 +1138,10 @@
     ;; makefile-*-mode
     (makefile-targets :foreground blue)
 
+    ;; man-mode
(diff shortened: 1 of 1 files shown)


ema2159:
Perfect!

ema2159:
Given that orange/green are a little bit too specific and didn't look pretty nice in other themes that are not Solarized I changed the `overstrike` `and0` underline colors to `operators` and `keywords` respectively (which keeps it the same for Solarized) so it looks better on other themes. Hope it's ok.

PR #327, opened 2019-08-27
Add doom-ewal-theme.

jjzmajic:


jjzmajic:
Ah dang. wrong time arguments everywhere all of a sudden.

jjzmajic:
Should be good now

jjzmajic:
This works with any pywal colorscheme. That way the doom style is preserved in anyone's config. Any preferences for screenshots (i.e which colorscheme would you like to see)?

As for the dependency... Would it perhaps make sense to move it to the theme file itself? If you want to keep it dependency free though, I get that. Just lmk

ema2159:
Regarding the screenshots, it's better if they have a code editing buffer like a C file or something and a buffer that shows Solaire mode like Treemacs or Ivy.

The thing is that I personally am not allowed to make such change (adding a dependency). So it would be up to the owner of the repo.

hlissner:
I would prefer this be a separate package (that pulls in doom-themes), because doom-themes is a theme pack and this isn't actually a theme, but rather an adapter for ewal themes.

jjzmajic:
Aaand done! Let me tell you this things looks sexy and was damn easy to implement. thank you @hlissner for writing wonderfully portable code. I'll post some screenshots. Doom just got 200 new themes.

jjzmajic:
doom-ewal
This is the theme sporting the "sexy-material" colorscheme (dw, the agenda doesn't have anything personal). But any pywal colorscheme works. I'll make a gif later.

PR #331, opened 2019-09-04
Fix customization type of padded-modeline.

tsuu32:
About `doom-themes-padded-modeline` and other `padded-modeline` suffixed customizable variables, I think `:type '(or integer boolean)` should be `:type '(choice integer boolean)`.

Elisp manual: https://www.gnu.org/software/emacs/manual/html_node/elisp/Composite-Types.html#Composite-Types


hlissner:
Good catch! Thanks for the help!

PR #334, opened 2019-09-29
Fontify helm-moccur-buffer as a link.

jjzmajic:
This face was not previously themed and is used by helm-occur, helm-ag...


hlissner:
Thanks for the help!

PR #336, opened 2019-10-01
Mod: magit hunk heading face for city-lights

am:
Using city-lights for a while now and there is a face that bothers me which is the magit hunk heading. I've changed the face to use less intrusive colors, and it works better for me. 

What are your thoughts about this change? Here is a snapshot from the Magit log for illustration of how it looks, both highlight and normal states:

current:
Screen Shot 2019-10-01 at 5 53 38 PM

modified:
Screen Shot 2019-10-01 at 5 51 19 PM

### DIFF
--- a/themes/doom-city-lights-theme.el
+++ b/themes/doom-city-lights-theme.el
@@ -121,9 +121,6 @@ determine the exact padding."
 
    (doom-modeline-bar :background (if -modeline-bright modeline-bg highlight))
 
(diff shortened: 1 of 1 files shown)


ema2159:
I'm ok with this. You should move the section with and add the respective comment down in the major mode's section though

am:
@ema2159 thank you for taking the time to look into this. I've made the proposed changes.

Issue #240, opened 2018-11-13
Theme Request: Oceanic Next


beojan:
This theme: https://github.com/voronianski/oceanic-next-color-scheme.
There is an Emacs version, in case that's helpful: https://github.com/terry3/oceanic-theme

ema2159:
Done!

PR #337, opened 2019-10-05
WIP: Add: OceanicNext theme

juanwolf:
This is the theme I've been using for quite a while with vim (don't judge me) and thought I would implement it since #240 is open (and it's hacktoberfest :tada:). 

I am not quite sure how to test it properly but here a screenshot for multiple languages:

### DIFF
--- a/doom-themes.el
+++ b/doom-themes.el
@@ -35,6 +35,7 @@
 ;;   [X] `doom-nord' (added by fuxialexnder)
 ;;   [X] `doom-nord-light' (added by fuxialexnder)
 ;;   [X] `doom-nova' (added by bigardone)
+;;   [X] `doom-oceanic-next' (added by juanwolf)
 ;;   [X] `doom-opera' (added by jwintz)
 ;;   [X] `doom-opera-light' (added by jwintz)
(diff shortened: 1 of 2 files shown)


ema2159:
Awesome! Let me check it and test it first, but it looks pretty cool. This is based from https://github.com/voronianski/oceanic-next-color-scheme am I right?

juanwolf:
Yes!

However, I could not find how to change the colour for different "kind" of keywords. If you look at the example, "import" and "const" have a different colour. Not sure if that's possible to change that and respect this [config](https://github.com/voronianski/oceanic-next-color-scheme/blob/master/Oceanic%20Next.YAML-tmTheme) closer.

ema2159:
@juanwolf don't worry about that, those are editor specific problems, and can't be solved directly from the theme implementation. What matters the most is that the color pallette is correct and that there's no visibility problems within the faces.

juanwolf:
Thanks @ema2159, I did a quick check installing sublime text, the original theme and checked few files in different format. The colors were wrong for some stuff like the keywords and the code was not readable inside the markdown format. The last commit should fix it and now the theme looks like this:

ema2159:
Perfect! Merging right now

PR #340, opened 2019-10-08
Gruvbox: Remove background from org ellipses 

teesloane:
**Before**

Screenshot 2019-10-08 14 21 04

**After**
Screenshot 2019-10-08 14 21 18


ema2159:
Hi! isn't it better to simply remove the background property?

teesloane:
Hello! Just tried that and it worked too — would you prefer that instead?

ema2159:
Yes!

teesloane:
Done!

PR #348, opened 2019-10-24
Add doom-blend-bg and doom-blend-fg functions

gagbo:
These functions allow to tweak colors along the background<->foreground
axis instead of the black<->white one.

### DIFF
--- a/doom-themes.el
+++ b/doom-themes.el
@@ -285,7 +285,7 @@ float between 0 and 1)"
 
 If USE-ALT is set to t, blend with the 'fg-alt' color instead of the 'fg' color."
   (cond ((and color (symbolp color))
-         (doom-blend-fg color alpha))
+         (doom-blend-fg (doom-color color) alpha))
 
         ((listp color)
          (cl-loop for c in color collect (doom-blend-fg c alpha)))
@@ -300,7 +300,7 @@ If USE-ALT is set to t, blend with the 'fg-alt' color instead of the 'fg' color.
 
 If USE-ALT is set to t, blend with the 'bg-alt' color instead of the 'bg' color."
   (cond ((and color (symbolp color))
-         (doom-blend-bg color alpha))
+         (doom-blend-bg (doom-color color) alpha))
 
         ((listp color)
          (cl-loop for c in color collect (doom-blend-bg c alpha)))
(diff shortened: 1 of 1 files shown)


hlissner:
What do these two functions do that `(doom-blend COLOR 'bg ALPHA)` (or `(doom-blend 'bg COLOR ALPHA)`) can't?

Also, can't both of these functions be simplified to a call to `doom-blend`? e.g.

```elisp
;;;###autoload
(defun doom-blend-fg (color alpha &optional use-alt)
  "Blend a COLOR (a hexidecimal string) with the current theme's foreground by a coefficient ALPHA (a float between
0 and 1).
If USE-ALT is set to t, blend with the 'fg-alt' color instead of the 'fg' color."
  (doom-blend color
              (if use-alt (doom-color 'fg-alt) (doom-color 'fg))
              (- 1 alpha)))

;;;###autoload
(defun doom-blend-bg (color alpha &optional use-alt)
  "Blend a COLOR (a hexidecimal string) with the current theme's blackground by a coefficient ALPHA (a float between
0 and 1).
If USE-ALT is set to t, blend with the 'bg-alt' color instead of the 'bg' color."
  (doom-blend color
              (if use-alt (doom-color 'bg-alt) (doom-color 'bg))
              (- 1 alpha)))
```

gagbo:
I didn't think about just using `doom-blend`, even while copy-pasting the code to make it work :(

I started thinking about this function because I saw in themes that colors were almost always tweaked with `doom-darken` and `doom-lighten`, and I thought the api would be clearer if it was "towards bg" or "towards fg". But that's not a necessary change for sure.

hlissner:
> I started thinking about this function because I saw in themes that colors were almost always tweaked with doom-darken and doom-lighten, and I thought the api would be clearer if it was "towards bg" or "towards fg". But that's not a necessary change for sure.

Ah, good point. Perhaps instead of adding these two new functions, we could add a third argument to `doom-darken` and `doom-lighten`, to specify a base. Instead of `"#FFFFFF"` and `#000000`. Perhaps it would make sense to use `(doom-color 'fg)` and `(doom-color 'bg)` as a default, instead?

gagbo:
That's another possibility yeah, I was just afraid that a bunch of themes would look different and that it would make people notice.

gagbo:
Definitely not needed

PR #353, opened 2019-10-30
Fix/ci with colors macro

gagbo:
# Fix

CI is fixed by changing the variable that the `-with-colors!` macro referred to

# Tweaks

I added some descriptions to the tests from what I understood

And used the option to make it clear that we need to load the test-helper file to run tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ all:
 	@cask
 
 test:
-	@cask exec ert-runner
+	@cask exec ert-runner -l test/test-helper.el
 
 clean:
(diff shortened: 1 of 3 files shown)


gagbo:
The passing pipeline : https://travis-ci.org/hlissner/emacs-doom-themes/builds/604887930

gagbo:
I thought it might make it easier to follow internal changes, since there was also a variable for `-color-p`

hlissner:
Looks good to me. Thanks for your help!

PR #363, opened 2019-12-11
add material theme

### LINKED ISSUE #318: Add Material theme
https://github.com/Emiller88/emacs-doom-community-themes/pull/2/files

tam5:
Closes #318.

Adds initial material theme, which is mostly the same as palenight theme with default material background.

(Future plans - configurable accent color, other material varations)

### DIFF
--- a/themes/doom-material-theme.el
+++ b/themes/doom-material-theme.el
@@ -1,0 +1,153 @@
+;;; doom-material-theme.el --- inspired by Material Theme by equinusocio
+(require 'doom-themes)
+
+;;
+(defgroup doom-material-theme nil
+  "Options for doom-themes"
+  :group 'doom-themes)
+
+(defcustom doom-material-padded-modeline doom-themes-padded-modeline
+  "If non-nil, adds a 4px padding to the mode-line.
+Can be an integer to determine the exact padding."
+  :group 'doom-material-theme
+  :type '(choice integer boolean))
+
+;;
+(def-doom-theme doom-material
+  "A dark theme inspired by Material Theme by equinusocio"
+
+  ;; name        default   256       16
+  ((bg         '("#1C262B" nil       nil))
+   (bg-alt     '("#263238" nil       nil))
+   (base0      '("#171F24" "black"   "black"))
+   (base1      '("#1C262B" "#262626" "brightblack"))
+   (base2      '("#222D32" "#303030" "brightblack"))
+   (base3      '("#171F24" "#3a3a3a" "brightblack"))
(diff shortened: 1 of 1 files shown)


ema2159:
Could you upload a screenshot please?

tam5:
Screen Shot 2019-12-11 at 7 41 40 PM

ema2159:
Nice! I will test it as soon as possible.

ema2159:
Are you using `(solaire-mode-swap-bg)` or taking it into account? It seems that you're assigning the `bg` color to the `bg-alt`'s and vice versa. The darkest background color is usually supposed to be `bg-alt`.

tam5:
Actually @ema2159 I have two questions on that:

1. I did intentionally flip the colors, so that when solaire mode is turned off, the "correct" background is shown, which would be the same color as the editor BG. On the other hand, if solaire mode is on, the editor remains the same, but the extra buffers like treemacs get additional contrast.

2. Yes you are correct I should be using `solaire-mode-swap-bg` but I'm not sure where exactly that is supposed to go. I don't see in the palenight theme at all. It seems like the right place for it currently would actually be in here https://github.com/hlissner/doom-emacs/blob/develop/modules/ui/doom/config.el#L3-L23 somewhere. Anyway, please advise, thanks!

ema2159:
1- I understand. The thing is that when I load the theme, the dark background is loaded in the main buffers and the light background is loaded in the secondary buffers (treemacs, which-key etc), on the contrary of the screenshot you posted (having `solaire-swap-bg` active).

2-`solaire-swap-bg` is external from the package, it goes into your config, and should always be active when using solaire-mode with doom-themes, as advised by hlissner.

themes/doom-material-theme.el
```diff
@@ -0,0 +1,153 @@
...
+  :group 'doom-material-theme
+  :type '(choice integer boolean))
+
+;;
+(def-doom-theme doom-material
+  "A dark theme inspired by Material Theme by equinusocio"
+
+  ;; name        default   256       16
+  ((bg         '("#1C262B" nil       nil))
+   (bg-alt     '("#263238" nil       nil))
```
ema2159: Switch these two (bg and bg-alt). That's would solve the problem.

wpcarro:
Nice! Good work, everyone. Excited to use this theme.

ema2159:
@tam5 awesome work! This implementation has become one of my favorite themes! I use it a lot for web dev because it allows me to focus incredibly well.

PR #365, opened 2019-12-14
Acario/lsp ui sideline

gagbo:
Commit on top of #364 

Add lsp-ui related faces : 
- sidelines as "more bg" comments
- lsp-ui-peek "trigger" (the searched symbol) as yellow in the menu


hlissner:
Thanks again!

PR #366, opened 2019-12-14
Acario/bright modeline tweaks

gagbo:
Commit on top of #364 

Change the bright modeline option to go back to a slightly modified blue


hlissner:
Thanks for the help!

PR #367, opened 2019-12-25
WIP: Add doom-manegarm theme

kenranunderscore:
Manegarm is a dark theme with an autumn-inspired color palette.

I wanted to create a dark theme that uses a non-white or not-very-light color for its foreground. After finally having settled for a darker tone of green, the rest of the color palette was already mostly complete.
I think it might be the first dark doom theme that uses a "non-standard" foreground color.

The finalized color palette reminded me a lot of album covers of some bands that I used to listen to. The one that stuck out the most was the Månegarm album "Vredens Tid" - hence the name.

For `org-mode`, I used some of the colors in the palette that are not very present elsewhere. I just need the org headers to stand out a bit more.

Now I'm primarily looking for feedback. At the moment, I don't have that much time, so I'm "collecting" first :) Here are some points I'm not 100% happy with yet or that are still `TODO`, and where feedback would thus really help:
- Things I'm doing horrendously wrong: it's the first time I'm fiddling on something with emacs and I probably made all kinds of silly mistakes, especially w.r.t. how doom themes are supposed to work. I had to make lots of adjustments because the way `def-doom-theme` uses the default colors sometimes clashes with the "colourfulness" of the theme.
- The modeline background color
- ~~The colors for `rainbow-delimiters-mode`~~
- Colors for modes I don't use and didn't have the time to test yet, for instance `treemacs`, `helm`, ...

Here's how it looks on my small screen. Any feedback is welcome and appreciated!
manegarm_1

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ pointers. Additional theme and plugin support requests are welcome too.
 [juanwolf]: https://github.com/juanwolf
 [JongW]: https://github.com/JongW
 [jwintz]: https://github.com/jwintz
+[kenranunderscore]: https://github.com/kenranunderscore
 [Material Themes]: https://github.com/equinusocio/vsc-material-theme
 [Moonlight Theme]: https://github.com/atomiks/moonlight-vscode-theme
 [mode-line]: https://github.com/hlissner/.emacs.d/blob/master/modules/ui/doom-modeline/config.el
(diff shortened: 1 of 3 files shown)


ema2159:
Looks pretty cool and original! I'll give it a try, test it and tell you if there's any modifications to be done.

kenranunderscore:
I had a first stab at `helm`, but I'm not really good at using it. I made it look similar to how my `ivy` minibuffer looks right now. I also changed the `selection` color that I didn't use in the theme definition. It is used in stuff like `helm` though and for other kinds of selections.

The more I'm getting used to `def-doom-theme` the more places I see where I misunderstood it...

ema2159:
I think this one is ready, is it?

kenranunderscore:
@ema2159 Sorry for the delay, I accidentally killed my Linux system trying to configure my new keyboard. It should be ready now!

ema2159:
That sounds like a very Arch problem haha.

PR #371, opened 2020-01-10
Increase salience of `helm-source-header`

obar:
Small change, makes it more clear which source one is looking at in a helm selection.


hlissner:
Thanks for the help!

PR #372, opened 2020-01-11
Extend the ivy-current-match background highlight to full line width

ianpan870102:
By making use of the new `:extend` keyword, we can make sure that the background highlight of the current ivy match doesn't end abruptly with the last word.
I've tested this small addition on my machine and it works as intended. Please consider the pull request. Cheers!


hlissner:
Thanks for the help!

PR #376, opened 2020-01-21
Add org-link styling for doom-challenger-deep

jethrokuan:
I'm overall pleased with the challenger-deep theme, but the default styling for org-link isn't outstanding enough.

Before:
image

After:
image


jethrokuan:
Seems like an issue on Emacs 28, not loading some styles...

ema2159:
Nice catch!

PR #386, opened 2020-01-31
fix region background from nord4 -> nord2 

mpenet:
From https://www.nordtheme.com/docs/ports/emacs/configuration

> Region Highlight Style
> By default, Nord Emacs uses nord2 as background color for selected text in the editor to provide a sane default style.

It's way easier on the eye imho and matches the other nord themes packages.

Thanks for your work! I keep juggling between some of these themes.


ema2159:
By default, nord2 (#434C5E) is mapped to base4, so the color is right in the theme. If you change to the base2 color which simply is a slightly lighter shade than nord0 (bg) the region section becomes not visible.

PR #396, opened 2020-02-11
Add org-journal calendar-entry face for doom-challenger-deep

jethrokuan:
Gives it a nice orange, compared to the default dark-red:

image

### DIFF
--- a/themes/doom-challenger-deep-theme.el
+++ b/themes/doom-challenger-deep-theme.el
@@ -163,7 +163,12 @@ determine the exact padding."
    (solaire-org-hide-face :foreground hidden)
 
    ;; tooltip
-   (tooltip              :background base0 :foreground fg))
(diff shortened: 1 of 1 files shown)


hlissner:
I've added org-journal faces to the base theme. It uses the theme's definition for magenta and red, so it should look nicer. What do you think? Is that alright, or is orange better? I don't mind making it the exception for challenger-deep -- I just wanted a sane default for the other themes.

jethrokuan:
I think it kinda looks okay now:

image

although the orange seems nicer. I wouldn't mind it either way, so I'll leave it up to you.

ema2159:
@jethrokuan are you alright with it now, or do you still think the orange version is better?

jethrokuan:
I'm fine with it.

PR #398, opened 2020-02-14
Fix/acario 256 colors

gagbo:
Fixes to acario palettes, including 256 colors for terminal """greatness""" (I prefer my terminal emulators with truecolor support, hence the quotes).

I'm still not totally over, I create a PR so everyone knows changes are coming. Hopefully the commit history gives an idea of what I'm trying to do.


gagbo:
Working on fixing this (extract from Discord)
> In the screenshot : 
> - top is tty emacsclient connected to a gui emacs shown on bottom.
> - you can see in both buffers the theme currently loaded
> - in the top emacsclient you can see that the default face uses `(car (cdr (fg))` (expected because 256-color terminal), but uses `(car bg)`  for the background, hence the blue background

> Recipe : 
> 1) Start the GUI
>    Doom-acario-dark uses the true color values, so no issues there
> 2) Start a tty emacsclient process :
>   - connected to the server in the first GUI
>   - in Konsole (also reproduced in urxvt) terminal, which doesn't use truecolor ; this has been confirmed by calling M-x list-colors-display which showed color-XXX colors
>   Doom-acario-dark uses the 256-color value for the `(doom-color 'fg)` color, but the "true color" value of the `(doom-color 'bg)` and I don't get why (Hence the blue background on the screen shot)
> 3) Bonus round
>   `SPC-: (load-theme 'doom-acario-dark t) RET` in the GUI emacs in the bottom will start using `(car (cdr bg))`
 as the background color. I.e if I had color-233 for the background in the 256 slot, this call would trigger a undefined color : color-233 error since GUI emacs has no idea what those colors are

fg_256-bg_tc

gagbo:
The issue I had with was solaire-mode not actually using the correct `(min-colors X)` across different frames, as seen with @hlissner (I'll make an issue at solaire-mode I guess)

Ready for a merge here :) Acario-light still looks a little weird in 256-color but I'll work on that later

PR #400, opened 2020-02-17
notmuch: Allow one to see added/deleted tags on unread mail

obar:
In notmuch, there are faces used to indicate which tags are being
added and which are being removed in the search buffer (of which the
inbox is one). With no themes loaded, `notmuch-tag-added` applies a
green underline and `notmuch-tag-deleted` applies a red
strikethrough. That combines just fine with
`notmuch-search-unread-face`, which simply makes the text bold.

The old approach of making most of the text bold and using color to
show unread doesn't work with the face order on notmuch. A tag being
deleted, rather than showing up as red, is colored same base8 as the
rest of the tags.

This commit changes the unread emphasis to bold rather than base8.

To make bold stand out for unread lines, other text has to be changed
to normal weight.


hlissner:
Thanks for the help!

PR #403, opened 2020-02-17
helm: Make helm-source-header a colored bar

obar:
An improvement on 6946f220ce581243d0aa9ee892e772f44425e14a

### DIFF
--- a/doom-themes-base.el
+++ b/doom-themes-base.el
@@ -630,7 +630,9 @@
      (&dark  :distant-foreground highlight)
      (&light :distant-foreground base0))
     (helm-match :inherit 'bold :foreground highlight :distant-foreground base8)
-    (helm-source-header          :background base2 :foreground keywords :weight 'bold)
+    (helm-source-header          :background base1
+                                 :foreground (doom-blend constants bg 0.7)
+                                 :inverse-video t :weight 'bold)
     (helm-swoop-target-line-face :foreground highlight :inverse-video t)
     (helm-visible-mark           :inherit '(bold highlight))
     (helm-moccur-buffer          :inherit 'link)


hlissner:
Please provide before and after screenshots.

obar:
Screenshot from 2020-02-17 16-29-25 helm-source-header

obar:
I should mention this change brings things more inline with what a `helm-source-header` [normally looks like](https://emacs-helm.github.io/helm/#advanced-usage), delineating parts of the helm buffer.

obar:
Screenshot from 2020-02-25 00-56-12

The same change in other themes in case that's useful to see. I tried in a few other themes as well before settling on the blended use of the `constants` color.

hlissner:
I understand the readability argument, but I'm not a fan of the change. I chose the original color because I didn't like how helm looked. At the very least, I don't think I want this change for `doom-one`.

I'll visit some of the other themes and see which ones it suits and revisit this when I get home later today.

obar:
Hey @hlissner!

I have another approach that uses variants of one color rather than additional colors for distinguishing the `helm-source-header` and `helm-selection`. I've been using it for a few weeks and it works well for elements standing out.

The changed faces:

```
  '(helm-selection
    (&all :inherit 'bold :extend t :distant-foreground highlight)
    (&dark :background (doom-lighten (doom-darken selection 0.5) 0.1))
    (&light :background (doom-blend bg selection 0.7)))
  '(helm-source-header
    (&all :foreground selection :inverse-video t :weight 'bold)
    (&dark :background base8)
    (&light :background base0))
```

Some screenshots:

Screenshot from 2020-05-26 23-1

Screenshot from 2020-05-26 23-2

Screenshot from 2020-05-26 23-3

I've also tried with these two styles swapped—brighter background for `helm-selection` and more muted for `helm-source-header`—but it didn't work as nicely across all themes. There's too much of a color similarity between `selection` if used directly in `helm-selection` and `highlight` used for `helm-match`, you can see the legibility impact of this in the "before" screenshots.

More tasteful?

obar:
Obviously this automated action is to clear out issues that were resolved elsewhere, and ones that aren't that important.

At first glance this might seem like a change that falls into the second category.

I believe this is a true usability improvement as the headings for different helm sources should be a salient separation between different categories, rather than blending in with the list items. For single-source helm interfaces this clearly doesn't matter, but for a multi-source interface like `helm-mini` which one uses multiple times per hour (and others that are still regular but less frequent, like `helm-projectile` and `helm-apropos`) one wants to see the different sources as clearly distinct.

Maybe the exact aesthetics still need to be tweaked beyond [the second go](https://github.com/hlissner/emacs-doom-themes/pull/403#issuecomment-634763618) (I've since made `helm-selection` slightly less bright in my config*) but in some form this change should still be made.

---------

\* `(&light :background (doom-blend bg selection 0.7))`

hlissner (changes requested):
Sorry for the incredibly late response. I've had a lot going on and only recently have returned to my OSS projects. Would you mind rebasing this PR onto the HEAD of `master`? I still can't say I'm a huge fan of the change (I prefer the low-contrast colors), but at the end of the day I don't use helm, so if this can be rebased I'll happily merge it.

PR #418, opened 2020-03-04
Add theme rouge

JordanFaust:
This is my first attempt at fully porting a theme. This is in reference to https://github.com/hlissner/emacs-doom-themes/issues/417

NOTE: this was ported from the nord theme as a base. Would have likely been better to start from the material theme. There might be a few colors left over from the nord theme that I was never able to surface through my workflows.

Variations from VSCode theme:
* Constants are 'orange' in this rather then 'yellow'. I used ruby as a base and it stayed closer to the overall look since constants and module/class definitions use a shared base.

I have only tested this with a few language modes (ruby elisp go). None of them are 100% matches. This is largely due to VSCode having more fine grained control over 'faces' in its supported languages

#  Rouge Theme
image

### DIFF
--- a/themes/doom-rouge-theme.el
+++ b/themes/doom-rouge-theme.el
@@ -40,44 +40,44 @@ determine the exact padding."
    (base2      '("#151D2B" "#2e2e2e" "brightblack"  ))
    (base3      '("#172030" "#262626" "brightblack"  ))
    (base4      '("#5D636E" "#3f3f3f" "brightblack"  ))
-   (base5      '("#A2A6AC" "#64727d" "brightblack"  ))
-   (base6      '("#E8E9EA" "#6b6b6b" "brightblack"  ))
+   (base5      '("#64727d" "#64727d" "brightblack"  ))
+   (base6      '("#B16E75" "#6b6b6b" "brightblack"  ))
    (base7      '("#E8E9EB" "#979797" "brightblack"  ))
    (base8      '("#F0F4FC" "#dfdfdf" "white"        ))
-   (fg         '("#bbb"    "#bbb"    "white"        ))
+   (fg         '("#BBBBBB"    "#bbb"    "white"        ))
    (fg-alt     '("#E5E9F0" "#bfbfbf" "brightwhite"  ))
 
-   (grey       base4)
+   (grey       base5)
    (red        '("#c6797e" "#c6797e" "red"          ))
-   (light-red  '("#AD6A6E" "#C6878F" "red"          ))
+   (light-red  '("#DB6E8F" "#DB6E8F" "red"          ))
    (orange     '("#eabe9a" "#eabe9a" "brightred"    ))
-   (green      '("#969E92" "#ADB9A4" "green"        ))
-   (teal       '("#8FBCBB" "#44b9b1" "brightgreen"  ))
-   (yellow     '("#EBCB8B" "#ECBE7B" "yellow"       ))
-   (blue       '("#1E6378" "#51afef" "brightblue"   ))
-   (dark-blue  '("#91d1bd" "#91d1bd" "blue"         ))
-   (magenta    '("#4C4E78" "#b18bb1" "magenta"      ))
+   (green      '("#A3B09A" "#A3B9A4" "green"        ))
+   (teal       '("#7ea9a9" "#7ea9a9" "brightgreen"  ))
+   (yellow     '("#F7E3AF" "#F7E3AF" "yellow"       ))
+   (blue       '("#6e94b9" "#6e94b9" "brightblue"   ))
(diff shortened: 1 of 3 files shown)


ema2159:
Thanks! Just for you to know, sometimes is really hard if not impossible to completely match all the colors for all the languages and modes of other editors so don't worry much about that. 
For some reason, people have been pretty active with themes recently :sweat_smile: and I got a new job so It may take some time for me to review them all.
With that said, I'm really glad to see people taking initiative and porting new themes! The repo has grown considerably with lot of incredible, high quality themes.

ema2159:
I modified the palette a little bit in order to match the original. Also I made comments and keywords italic as the original.

ema2159:
Also, I'm adding a `centaur-tabs-plain-icons` custom to make `centaur-tabs` icons' color the same as the foreground to match the theme

themes/doom-rouge-theme.el
```diff
@@ -0,0 +1,182 @@
...
+   (selection      base4)
+   (builtin        light-red)
+   (comments       grey)
+   (doc-comments   green)
+   (constants      red)
+   (functions      salmon)
+   (keywords       magenta)
+   (methods        light-red)
+   (operators      magenta)
+   (type           magenta)
```
JordanFaust: This is something that I explicitly made different. This is due to the difference in markup of code from VSCode to Emacs. This might be something that just needs to be overwritten on a language bases. But this significantly shifts from what it looks like for Ruby when comparing the themes side by side

JordanFaust: Not sure if we want to go down that route but in keeping with the 'spirit' of the theme I think keeping type faces red reflects more closely to the theme in VS Code

ema2159: The thing is, the theme was designed specifically for web dev (JS, HTML, CSS). If you check the repo, the only language specific customization is for those languages. All the other (C, Python, Ruby...) fall into the defaults I think, so what I did was to grab specifically the JS colors. By the way, you now may try centaur-tabs plain icons custom to make the theme even better.

JordanFaust: Do you have a code example where this is diverging for js/html when type is set back to red/orange compared to magenta. VSCode technically uses both the red/orange/yellow to highlight different identifiers/types in the code. But in emacs we lose most of that. Setting the type to red brought more of it into the theme. Copying exact matches for specific kinds of 'faces' from VSCode might be technically right but it looks significantly different when comparing the two side  by side.

Most of the improvements on color choice here are great. This is is just something that I think we will need to make our own for the conversion to an emacs theme. We lose out on a ton of different coloring due to the difference in syntax highlighting. And the red is the center piece of the whole theme.

ema2159: I would need to look into it. Could you show me an screenshot contrasting the two options in order to be in the same page?

JordanFaust: Here is JS example. Javascript looks nothing like the vscode theme because of the lack of face modification for normal variables.

image

Here is ruby with the change you made
image

As you can see the type definition being sent to magenta is a drastic shift away from the vscode theme. Here is the same with 'type' set to red. Still not a 100% representation of the theme but closer then the above

image

ema2159: The problem comes with things like this one:
image

If we changed the theme as you say, both the const and the var words would be red in JS. The problem is also that the keywords are magenta. The problem you mention is also present in Python, and it's unfortunately a pretty common problem.

A good idea for a package would be a package to extend the syntax highlighting of some modes. There's some packages that do that like js2-mode or web-mode for example.

JordanFaust: Is that screenshot from emacs? My comparison with my changes in does not make any changes to what I see within emacs (I'm using Doom and have the javascript module enabled and using the js2-mode). In fact my suggestion of the change on type does not influence that code snippet at all. 

image

I feel like outside of some narrow places this is a hard theme to port without some modifications. Because half of the languages don't even come close to the syntax highlighting available in VS Code that this theme leans so heavily on.

ema2159: It seems you're right. It doesn't affect Javascript. It will affect other modes such as C though.

ema2159: By the way, you can do `(setq centaur-tabs-plain-icons t)` to make the icons the same colors as the tabs' text, which will make it look more similar to the original.

ema2159:
@JordanFaust Merged. Excellent work brother. I did the changes that you requested, although it was in the tradeoff of the syntax highlighting of some other strongly typed languages like C or C++. It would be great if Emacs offered better customization options per major mode though.

JordanFaust:
@ema2159 Thanks! I appreciate all the help on this!

PR #439, opened 2020-04-15
Change background of selection

laishulu:
base4 is hard to distinguish from default background.
use base5 instead

screenshot for base4:
图片

screenshot for base5:
图片


hlissner:
Thanks!

PR #440, opened 2020-04-16
Fix fixed-pitch-serif face

x-ji:
I encountered an error while trying out the themes.

Seems that the face should be `fixed-pitch-serif` instead of `fixed-pitch-serif-face`.

Only the file `doom-acario-light-theme.el` has `fixed-pitch-serif` instead of `fixed-pitch-serif-face`, and I was able to use the theme successfully.


hlissner:
Thanks for the help!

PR #451, opened 2020-05-01
Minor face improvements for doom-one and doom-gruvbox

WhatDothLife:
## doom-one
Since latex' math-mode and keywords share the same color, I changed the one of math-mode to green (which IMO looks pretty neat).
onelatex
onelatex2

## doom-gruvbox
So far quoted elisp symbols and function names shared the same color. I made use of dark-cyan to make them distinguishable.
gruvbox1
gruvelisp2

The headers of a mail opened in view mode look pretty uniform. I spiced up the keys with red, to make it more readable.
gruvmu4e
gruvmu4e2

This one is a bit more opinionated. Latex' math-mode uses blue as its face. I think it doesn't really complement the pastel look of the theme, and thus changed it  to dark-cyan.
gruvlatex1
gruvlatex2

### DIFF
--- a/themes/doom-gruvbox-theme.el
+++ b/themes/doom-gruvbox-theme.el
@@ -91,7 +91,7 @@ determine the exact padding."
    (org-quote `(,(doom-lighten (car bg) 0.05) "#1f1f1f")))
 
   ;; --- extra faces ------------------------
-(
+  (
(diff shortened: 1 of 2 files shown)


hlissner:
I think those look good and make sense! Thanks!

PR #452, opened 2020-05-03
Enhance coloring of org-files

WhatDothLife:
So far, the coloring of org-files has been biased towards a heavy use of yellow. Overall, I didn't find its design very appealing or sometimes even misleading. This pull request is a try to fix these issues. The coloring is in parts inspired by doom one.

### Design guidelines
+ Don't mash up too many colors.
+ Gruvbox' yellow and orange are very sharp colors, scale them back in favor of e.g. violet.
+ Red is one of gruvbox' trade mark colors, yet signals errors or urgent tasks. Try to integrate it unobtrusively.
+ TODO keywords should be distinguishable from each other.
+ The overview visibility should already look neat.
+ Consider documentation as well as personal organization files (with timestamps, drawers etc.)

### Screenshots
The results look like the following (old version first):

Sample org file:
gruvbox_sample_old
gruvbox_sample_new

Overview:
gruvbox_lsp_overview_old
gruvbox_lsp_overview_new

lsp documentation:
gruvbox_lsp_old3
gruvbox_lsp_new3

Of course I'm open for discussion :)


midchildan:
Please take what I say with a grain of salt, but the fix could use more warmer colors. I think the current proposal is a bit too heavy on purple, making it look like the one dark theme instead.

WhatDothLife:
You definitely got a point there. I think purple for level 1 headings looks fine, but turning org-links to purple might have been a bit overkill. They could be reverted to yellow, which makes the TOC very dominant, but that's ok then. Dates could be reverted to yellow, maybe also orange... I chose tags-face to have the color of comments, but that was also a rather personal choice. Any proposals?

midchildan:
My personal preference would be to mimic how the original gruvbox theme looks with markdown. One thing I noticed looking at the original version is that doom-gruvbox's yellow stands out more despite both using yellow for headings. I think the primary reason for this has to do more with how org mode is used, rather than the choice of color itself. Because we constantly fold sections with org mode, we would often see a large chunk of yellow dominating the buffer. So I guess as long as we don't use yellow as a top or second level heading, things would look a lot better.

Here's how the original gruvbox looks with markdown BTW:
gruvbox

ema2159:
I'm not quite sure with this one. I mean, Gruvbox is an "earthy" theme, so I think that the use of yellow is preferred over violet since it's a more earthy color. 
Maybe you're right when you say we may be overusing it, but although I would be open to change the theme in order to use some other colors where yellow is being used, I think that yellow is ok as the "main color", and I would keep the heading 1 color yellow.
That being said, introducing some green, orange or red would be alright.

P.D. Sorry for the late response, I've been quite busy in several dimensions of my life these past few weeks.

hlissner (changes requested):
Sorry for the tremendously late response, but I'd like to put this PR to rest so let's go with this:

1. Please resolve the merge conflict by rebasing onto `master` and
2. Use violet for first level org headings, but purple for links. (The yellow also clashes with TODO keywords, so I think it's best we avoid it)
3. The other color changes are fine.

Besides, I think this is on the scale where it can be trivially changed by end users with `custom-set-faces` or similar.

hlissner:
I made the requested changes in cd1baec. In any case, thanks for bringing this to my attention!

PR #455, opened 2020-05-07
Add deadgrep-match-face to monokai pro and spectrum themes

daveduthie:
I'm not sure if a PR like this is acceptable, given that [deadgrep](https://github.com/Wilfred/deadgrep) isn't included in doom emacs by default, but I'm trying my luck!

P.S. deadgrep is awesome! I'd be keen to write a module for it.

### DIFF
--- a/themes/doom-monokai-pro-theme.el
+++ b/themes/doom-monokai-pro-theme.el
@@ -85,6 +85,9 @@ Can be an integer to determine the exact padding."
    (lazy-highlight                               :inherit 'match)
    (isearch-fail                                 :foreground red)
 
+   ;; deadgrep
+   (deadgrep-match-face                          :inherit 'match :box `(:line-width 2 :color ,yellow))
+
    ;; swiper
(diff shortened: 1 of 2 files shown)


ema2159:
It is indeed acceptable. But I think it would be a good idea to add it not just to Monokai pro, but to the base file.

daveduthie:
Added `deadgrep-match-face` to the spectrum variant too.

@ema2159, the reason I haven't added to the base theme is that `deadgrep-match-face` inherits from `match`, which is overridden in these two monokai themes (pro and spectrum), but not in any other themes.

As a result, deadgrep matches were only broken for these two themes.

Deadgrep matches in `doom-one`

image

Deadgrep matches in `monokai-pro` (without patch applied)

image

Deadgrep matches in `monokai-pro` (patch applied)

image

Let me know if this makes sense?

ema2159:
Perfect. Thanks for your help!

PR #457, opened 2020-05-08
Fix internal README links for one-light & vibrant

ilkermutlu:
Links for doom-one-light and doom-vibrant had a trailing '-wip' whereas the targets didn't.


hlissner:
Good catch. Thanks for the help!

PR #471, opened 2020-05-21
Change constant & type colours according to official screenshots

ianpan870102:
According to the official gallery here https://github.com/morhetz/gruvbox/wiki/Gallery, constants should be gruvbox-violet, and types should be gruvbox-yellow.


ianpan870102:
This PR fixes this issue: https://github.com/hlissner/emacs-doom-themes/issues/472

PR #491, opened 2020-06-30
Override helm settings in dracula-theme

kngwyu:
Thank you for this nice package!
This PR overrides some colors of `helm` using the settings from [the original dracula theme for emacs](https://github.com/dracula/emacs).

Before:
image
After:
image

Note that I didn't override `helm-selection` and `helm-line-selection` since I like this blue one.


themes/doom-dracula-theme.el
```diff
@@ -237,6 +237,29 @@ determine the exact padding."
...
+   (helm-buffer-size :foreground fg)
+   (helm-candidate-number :foreground bg :background fg)
+   (helm-ff-directory :foreground green :weight 'bold)
+   (helm-ff-executable :foreground dark-blue :inherit 'italic)
+   (helm-ff-invalid-symlink :foreground magenta :weight 'bold)
+   (helm-ff-prefix :foreground bg :background magenta)
+   (helm-ff-symlink :foreground magenta :weight 'bold)
+   (helm-grep-finish :foreground base2)
+   (helm-grep-running :foreground green)
+   (helm-header :foreground base2 :underline nil :box nil)
```
hlissner, line 254: Could you restore the background of `helm-source-header`? It is intended to make it more distinct from the other entries.

```diff
-(helm-source-header :foreground base2 :underline nil :box nil)
+((helm-source-header &override) :foreground base2 :underline nil :box nil)
```

The `&override` will cause it to inherit the properties from the base theme.

kngwyu: Thanks, I changed it to only override the `foreground`.

ema2159:
Is this ready to go?

PR #493, opened 2020-07-18
Add:  doom-old-hope-theme.el

teesloane:
Based off of [An Old Hope Syntax](https://github.com/jesseleite/an-old-hope-syntax-atom)

Screens demonstrating a mix of different languages:

Screen Shot 2020-07-17 at 10 58 26 AM
Screen Shot 2020-07-17 at 8 05 29 AM
Screen Shot 2020-07-17 at 11 00 22 AM

### DIFF
--- a/themes/doom-old-hope-theme.el
+++ b/themes/doom-old-hope-theme.el
@@ -34,7 +34,7 @@ determine the exact padding."
 
   ;; name        default   256       16
   ((bg         '("#1c1d20" "#1c1d20"       nil))
-   (bg-alt     '("#313339" "#313339"       nil))
+   (bg-alt     '("#151619" "#151619" nil))
    (base0      '("#1B2229" "black"   "black"))
    (base1      '("#1c1f24" "#1e1e1e" "brightblack"))
    (base2      '("#202328" "#2e2e2e" "brightblack"))
@@ -65,7 +65,7 @@ determine the exact padding."
    (vertical-bar   (doom-darken base4 0.2))
    (selection      red)
    (builtin        yellow)
-   (comments       base4)
+   (comments       base5)
    (doc-comments   (doom-lighten blue 0.25))
    (constants      orange)
    (functions      yellow)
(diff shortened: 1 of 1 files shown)


ema2159:
Gorgeous! I can't wait to review this!

ema2159:
Pretty awesome theme. I love its contrast. I just have a couple remarks:
1-The bg-alt color should be `(bg-alt     '("#151619" "#151619"       nil))`. It's likely that you missed this because you're not using solaire-mode (which you should! It drastically improves the looks of your Emacs and all Doom themes support it).

2-I don't think that the background from the `ivy-current-match` face should be removed. First, it's hard to follow if there's no background, and second, all Doom themes have background for that face, so it should be kept like that to remain consistent. Finally, the face is duplicated in lines 150 and 187, with the second one being the one that takes effect, so please remove the first one!

teesloane:
Thanks for the feedback! I've made the requested changes as well as some small tweaks to org mode since giving this a test drive.

ema2159:
Awesome. I just have a final question. Which other editor you used for comparison and which major mode? The thing is, it is too hard to exactly match a theme in Emacs to its implementation in another editor due to the lack of individual customization for each major mode, so usually what I do, and what I advise to do its to select a major mode of a popular language (Python, Javascript, C etc.) and use that as a reference. For example:
image
Here you can see your port, and VS Code's implementation both showing the same Python file. You can see that both of them have differences between the highlighting of the different types of keywords, which is completely fine, the palette is there. So if Python's implementation looks a little bit different because you matched the colors in lets say Javascript, that's completely fine, but if not, then we can choose one of the major modes and tailor the theme there.

teesloane:
Yep, good point. I matched against JS and Clojure in VsCode / emacs - the colours applied differently, but I'm okay with the palette at least 🤷 .

ema2159:
Yes! The only color that bugs me is the keywords color. Is there a particular reason for it to be green instead of red? Otherwise, its ready to be merged.

teesloane:
Yeah, the red was a bit overbearing in some languages (Elm, if I remember correctly). No worries though - if it's too strong, I'll mod my local version :). Thanks for the review and feedback.

ema2159:
Thank you very much for this theme! It's an incredible addition. Regarding the colors issues, yeah. It's unfortunate that Emacs doesn't offer a more granular control over its major mode theming. I've sacrificed many times the look of one major mode for another, which is really not ideal.

PR #496, opened 2020-07-23
[doom-base] Add web-mode support for json-keys and context

ianpan870102:
This small commit targeting doom-base makes sure that all **web-mode json-keys and json-context will be styled as strings** (following the design guideline used by Visual Studio Code).


hlissner:
Makes sense to me. Thanks!

PR #507, opened 2020-08-09
Add doom-flatwhite theme

ShaneKilkelly:
This is a port of [flatwhite-syntax](https://github.com/biletskyy/flatwhite-syntax), by Dmitry Biletskyy.

## Screenshots

Dashboard:

image

With an Ivy menu:

image

Which Key:

image

Python:

image

With Swiper:

image

Elisp:

image

Javascript:

image

Clojure:

image

Racket:

image

Ruby:

image

Dired:

image

Markdown:

image

Magit:

image

String search (`int`, in this case):

image

## Review Notes

- This theme goes a little against-the-grain, as it colors the background behind the text to achieve the desired highlighting. As a result, this theme file is quite a bit longer than most others, with more explicit face definitions. 
- All the modes I've tested work pretty well, but I'm open to feedback on particular choices made.

### DIFF
--- a/themes/doom-flatwhite-theme.el
+++ b/themes/doom-flatwhite-theme.el
@@ -141,17 +141,16 @@ determine the exact padding."
 
     (default :background bg :foreground fg)
 
-    (font-lock-builtin-face              :inherit '(italic default) :extend t)
-    (font-lock-comment-face
-      :foreground comments
-      :background  bg)
+    (font-lock-builtin-face              :foreground fg
+                                         :inherit 'italic :extend t)
+    (font-lock-comment-face              :foreground comments)
     (font-lock-comment-delimiter-face    :inherit font-lock-comment-face)
     (font-lock-doc-face
       :inherit 'font-lock-comment-face
       :foreground doc-comments
       :slant 'italic)
     (font-lock-type-face                 :inherit 'default)
-    (font-lock-variable-name-face        :inherit 'default)
+    (font-lock-variable-name-face        :foreground fg)
     (font-lock-warning-face              :background fw-red-blend
                                          :foreground fw-red-text)
     (font-lock-negation-char-face        :inherit 'default)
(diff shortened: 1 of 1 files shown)


ema2159:
That's quite unique!

Emiller88:
@ShaneKilkelly Great work! One thing I noticed was some weird highlighting of `int` in the middle of `println` in your java screenshot.

Second I think the org-mode src block line color is a little harsh compared to the markdown ones
image

ShaneKilkelly:
Thanks @Emiller88 ! I'll push a commit to fix the org src block problem. In addition, I'll go over the org styles in more detail, and try to get it all to look good.

On the multi-line strings, I've looked into it, and I think this is an unfortunate side-effect of coloring the _background_ , rather than the foreground, as most themes do. I agree it does look a bit ugly in these cases, but from what I can tell, emacs just thinks it's all `font-lock-string-face`, and doesn't distinguish between "proper" string content and the accidental whitespace that comes ahead of multiline strings.

ShaneKilkelly:
@Emiller88 I think this might work well for org source blocks, what do you think?
image

Emiller88:
@ShaneKilkelly 
> unfortunate side-effect of coloring the background

Yeah that's the sad conclusion I came too also. But oh well, there's not many times I have a multiline string, I'll just have to figure out a better way to do syntax highlighting in my package `nextflow-mode`, since they're more of src blocks than strings.

I think those look better than the reverse oreo look! But quotes still have it also:
image

ShaneKilkelly:
@Emiller88 thanks, I'll fix the quotes too

Emiller88:
@ShaneKilkelly I also noticed that the current line highlighting removes the syntax highlighting
image

ema2159:
@Emiller88 that's expected. Maybe an option would be bolding the characters instead of highlighting their background.

ShaneKilkelly:
@ema2159 I like that idea, as it would still give you some kind of hint as to the syntax. I'll try it out :) thanks

ema2159:
I think it's ready. Merging now. Great job @ShaneKilkelly!

ema2159:
@ShaneKilkelly by the way... are you the author of Overleaf?... If yes well, thank you for making my time at the university one million times easier :smiley:. Such a great tool, literally everybody in my engineering school uses it.

ShaneKilkelly:
@ema2159 thanks :) 

I'm one of the original ShareLaTeX team, and now at Overleaf. I'm glad the tool has been of use to you!

PR #511, opened 2020-08-20
Update nord color to use a more true background

C0DK:
The background color used should be for background not not the editor pane it self. much like the dark version of the website https://www.nordtheme.com/docs/colors-and-palettes

Ideally the background color should/could be used for the line numbers - that would look good and make the main editor "pop" / seem elivated.

### DIFF
--- a/themes/doom-nord-theme.el
+++ b/themes/doom-nord-theme.el
@@ -40,7 +40,7 @@ determine the exact padding."
   "A dark theme inspired by Nord."
 
   ;; name        default   256       16
-  ((bg         '("#2E3440" nil       nil            ))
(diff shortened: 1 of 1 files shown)


hlissner:
Please provide before and after screenshots.

ema2159:
This is most likely a solaire thing, since #2E3440 is indeed the correct background from the palette.

ema2159:
I'll be closing this for now, but feel free to ask if you need further assistance.

C0DK:
Sorry yes i realize it was the correct colors. it was something with me upgrading that broke the solaire thing, where my main buffer wasn't colored. Thank you for great support.

PR #513, opened 2020-08-25
New theme doom-one-dark

kaue:
This is an alternative version from doom-one but with black background. currently my favorite theme :)


hlissner:
Thanks for the PR, but I won't accept themes that are just slight tweaks of other themes. If you just want to change one face, use `doom-one` and put this in your config:

```elisp
(custom-set-faces
  `(default :background "#000000"))
```

PR #514, opened 2020-09-03
[flatwhite] green string highlighting for web-mode

CarlQLange:


ema2159:
Please upload some before and after screenshots!

CarlQLange:
Sorry, of course! I did not pay any attention while I was doing this! 

Before:
Screenshot 2020-09-03 at 23 18 17


Here's an after: 
Screenshot 2020-09-03 at 23 17 05

You can see that the strings are really lost in the before screenshot. Furthermore, from a very quick google, it seems like the green background is how it's implemented for flatwhite on other platforms:

CarlQLange:
Looking at the final screenshot of the sublime text implementation I guess the `web-mode-variable-name-face` should also have a blue background.

CarlQLange:
Ah, wrong colour, should have been `blue`, not `teal`, but you get the picture

ShaneKilkelly:
Huh, I could have sworn I checked web mode when porting this theme :) anyway, good catch

ShaneKilkelly:
I could take a look at the variable-name thing and check if it matches my understanding of how the original theme is supposed to work.

ShaneKilkelly:
So, I think the blue highlight on variables might be unique to the Sublime Text implementation. The original Atom implementation does not seem to highlight variables in this way (https://github.com/biletskyy/flatwhite-syntax). 

However, the blue highlight does look pretty cool in some ways... so... I wonder if we could add this as a configurable option for the theme?

ema2159:
@ShaneKilkelly that is an option, but I'm fine with adding that blue as-is. Flatwhite syntax is very subtle, so even with that color it is not overwhelming.

CarlQLange:
Personally I think the more background highlights, the better. Flatwhite's whole style is based around those backgrounds and I think they look really awesome.

ShaneKilkelly:
Alternatively, could we add the blue highlight by default, with an option to turn it off? 

The flatwhite concept seems to have been built around a minimalist aesthetic, which can certainly be divisive. I think I'm leaning toward agreement with the consensus that more highlights are good in the general case, but I would also like the option to stick closer to the original concept. 

I can open a separate PR to add the blue highlight for vars, if we're all in agreement? :)

ema2159:
@ShaneKilkelly that's fine! You can add it with a defcustom like some configurations in other themes.

ShaneKilkelly:
@ema2159 done :) https://github.com/hlissner/emacs-doom-themes/pull/518

PR #518, opened 2020-09-16
Flatwhite: Add blue highlight to variable names

ShaneKilkelly:
As discussed in https://github.com/hlissner/emacs-doom-themes/pull/514

The original [flatwhite-syntax](https://github.com/biletskyy/flatwhite-syntax) theme for Atom does not highlight variable names. However, the [Sublime Text](https://github.com/Willamin/flatwhite) version of the theme does highlight variable names in blue.

This change adds a blue highlight to variable names, with a config variable (`doom-flatwhite-no-highlight-variables`) to turn the highlight off, for those who prefer the original theme.

This also changes the highlight color for keywords in `racket` and `clojure` (for example: `#:stuff` and `:stuff` respectively) from blue to orange, to better fit with the new usage of blue as the variable name color.

### After

flatwhite-var-highlight

### Before

Or, with `(setq doom-flatwhite-no-highlight-variables t)`

flatwhite-no-var-highlight

### DIFF
--- a/themes/doom-flatwhite-theme.el
+++ b/themes/doom-flatwhite-theme.el
@@ -453,14 +453,14 @@ determine the exact padding."
     (js2-external-variable :foreground fg)
 
     ;; racket
-    (racket-keyword-argument-face :foreground fw-blue-text
-                                  :background fw-blue-blend)
+    (racket-keyword-argument-face :foreground fw-orange-text
+                                  :background fw-orange-blend)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Sorry for the late response, but thanks for the PR! I like this idea; frankly I've wanted differentiated variables to be something all of our themes do -- even if it deviates from the themes' original specs.

Issue #422, opened 2020-03-19
Light theme background not loaded in `emacs -nw`


zakkak:
This is a duplicate of https://github.com/hlissner/doom-emacs/issues/2392 I created to discuss it on the correct repository.

As the original issue states, when using a light theme and start emacs in a terminal (e.g. `emacs -nw`) the background of the theme doesn't get loaded.
This is on purpose in order to allow transparency in emacs for transparent terminals. 

I think this should be made toggleable though, since some of us like our terminal dark and transparent but prefer our emacs light and opaque.
Could someone please direct me on how to implement this?

zakkak:
Currently I am using the following workaround, which enables a dark theme on the terminal:

```
(if (display-graphic-p)
    (progn
    ;; if graphic
      (setq doom-theme 'doom-solarized-light))
    ;; else (optional)
  (setq doom-theme 'doom-molokai))
```

YassinRian:
had the same problem ..this is what I did 
add the following to init file : (custom-set-faces! '(default :background "white")) 
extra addition to config file : source (https://mirrors.nju.edu.cn/elpa/melpa/doom-themes-readme.txt)   
(require 'doom-themes)

  ;; Global settings (defaults)
  (setq doom-themes-enable-bold t    ; if nil, bold is universally disabled
        doom-themes-enable-italic t) ; if nil, italics is universally disabled

  ;; Load the theme (doom-one, doom-molokai, etc); keep in mind that each
  ;; theme may have their own settings.
  (load-theme 'doom-one-light t)

  ;; Enable flashing mode-line on errors
  (doom-themes-visual-bell-config)

  ;; Enable custom neotree theme
  (doom-themes-neotree-config)  ; all-the-icons fonts must be installed!

kuba-orlik:
I'm experiencing this issue, as well. A flag to disable such behaviour would be very useful!

hlissner:
#519 was merged, which should resolve this issue.

In any case, what's happening here is that many Doom themes _do not set a background for their 256 variants_. This is intentional, because I expected folks to want to use their terminal backgrounds, rather than Emacs casting a secondary background on top of it.

Over time, I'll add backgrounds to each of our themes to deal with this (it's easier to turn off than on). In the meantime, you can specify an explicit background by adding this to your config:

```lisp
(custom-set-faces '(default ((t (:background "red")))))
```

If you are using Doom, then:

```lisp
(custom-set-faces! '(default :background "red"))
```

[...]

PR #519, opened 2020-09-28
Make the solarized-theme actually light in 256 colors mode

kuba-orlik:
This addresses the issue mentioned in #422 

It seems that for some reason the "256 colors" variant of solarized theme was just a dark b&w theme and rendered dark text on dark background. This should fix that, by using the same colors in 256 mode as in regular gui mode


hlissner:
Thanks for the help. I'd much prefer those colors were scaled down to the 8-bit space, but this will do for now. I may remove the background colors however. They are intentionally transparent, so that Emacs uses the background of your terminal instead of casting its own background on top of it, but I'll figure that out later. Thanks again!

Issue #503, opened 2020-07-30
[outrun-electric] Color mismatch on Emacs 26.3 (nox)


apiraino:
Hi,

I'm trying to figure out this problem. None of the themes I've tested (outrun-electric, laserwave, acario-dark) appear on my Emacs as I see on the screenshot.

Example: the background color for 256 colors terminals is `#0c0a20` but what I see is `#00005f` (which appears to be `color-17` when I look at `M-x list-colors-display`). All the colors are shifted in a way I don't understand and [don't match the screenshot](https://github.com/hlissner/emacs-doom-themes/blob/screenshots/doom-outrun-electric.png):

![2020-07-31_003858](https://user-images.githubusercontent.com/6098822/88981452-44ed9580-d2c6-11ea-8846-b2bcc5443dff.png)

Some more info:
- Gnome-Terminal with some color customizations (but I think this is not relevant)
- Using Sway/Wayland (but same issue appears also on i3/X11)
- Emacs 26.3 nox
- `echo $TERM` => `xterm-256color`
- I'm using just one theme ("use more theme" option is not checked)
- tried using other terminals (`st`, `kitty`, ...) and colors change in different ways

I believe it's an issue on my end but I can't make sense of all these info I've gathered.

Why some themes use the emacs color names instead of the RGB triplet? (example [acario-dark](https://github.com/hlissner/emacs-doom-themes/blob/180747a4cb848b6378b13bcaf908ea12c95bb4a6/themes/doom-acario-dark-theme.el#L37))

Any suggestion is greatly appreciated - thanks!!

Also, these themes are absolutely gorgeous.

hlissner:
Are you using solaire-mode? If so, turn it off. It doesn't work in the terminal.

apiraino:
I've checked, it's not installed

ema2159:
Is that terminal Emacs? There are some themes that currently doesn't support term Emacs that good. This is the case for Outrun.

apiraino:
@ema2159 you raise a very good point and this is why I explicitely mentioned that I am using `emacs-nox` (no X11), `M-x list-colors-display` lists 256 colors.
I tried installing `emacs-gtk` and `emacs-lucid` on another Ubuntu and the previous command returns ~550 colors (but the theme colors are still off from the screenshot).

So my question is: what are the requirements in order to attain the closest result as possible to the screenshot? Should I install your full Emacs configuration?

apiraino:
[...]

PR #522, opened 2020-10-03
How to enable 24bit colors on a non graphic shell

apiraino:
As suggested in #503 here's a bit of blurb to help emacs users without GUI compiled in.

How does it look? Please let me know how can I improve the wording.

p.s. Emacs did its thing removing some stray spaces
p.p.s. I'm not here for the tshirt :smile:


hlissner:
Sorry for the late response, but thanks for the help!

PR #525, opened 2020-10-16
doom-gruvbox-light: fix cursor bug, remove line numbers background

drchsl:
### Cursor fix and new option
- fixes the cursor being white after startup and black after reloading the theme 
- cursor is now grey, as in the [original gruvbox light](https://github.com/morhetz/gruvbox#light-mode) 
~~- the black cursor can be brought back by setting `doom-gruvbox-light-darker-cursor` to `t`~~

### Further port improvement

- in the [original](https://github.com/morhetz/gruvbox#light-mode) version, line numbers have no special background

### DIFF
--- a/themes/doom-gruvbox-light-theme.el
+++ b/themes/doom-gruvbox-light-theme.el
@@ -146,14 +146,14 @@ background contrast. All other values default to \"medium\"."
     :foreground doc-comments
     :slant 'italic)
 
-   (cursor    :foreground fg)
+   (cursor :background base4)
 
    ;; Line number
-   ((line-number &override) :foreground base4 :background base1)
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Could you remove the `doom-gruvbox-light-darker-cursor` option? Unless the implementation is particularly complex, I want to avoid adding options for trivial differences (unless the same/similar option exists upstream). Folks can use `custom-set-faces` otherwise.

drchsl:
Ah, totally makes sense. Removed it. I only thought "folks might have gotten used to that black cursor" and forgot how easy it is to change that.

hlissner (approved):

hlissner:
Thanks!

PR #526, opened 2020-10-17
Change org-ref ref face to teal

tecosaur:
Because red is the error/invalid link face, and I don't like thinking my perfectly-fine references are invalid.


hlissner:
In what them is it invalid? Red should be defined in all themes.

tecosaur:
Oh, no - I don't mean that the colour `red` isn't valid in a theme, I mean that the styling of `org-ref-ref-face` is currently the same as that of an invalid link in org, like `[[/file/that/does/not/exist]]` - which doesn't seem good to me.

hlissner:
Ah, gotcha. Thanks!

PR #541, opened 2020-11-21
solarized-dark: change variables to blue

sulami:
This is in line with solarized-light, which is more readable imho.

### DIFF
--- a/themes/doom-solarized-dark-theme.el
+++ b/themes/doom-solarized-dark-theme.el
@@ -81,7 +81,7 @@ determine the exact padding."
    (operators      orange)
    (type           yellow)
    (strings        cyan)
-   (variables      fg)
+   (variables      blue)
    (numbers        magenta)
    (region         base0)
    (error          red)


hlissner:
Could you include a before and after screenshot?

sulami:
Before and after, in YAML, where it annoyed me to begin with:

before and after

Solarized light for reference:

solarized light

hlissner (approved):

ema2159:
I've been thinking about this for quite a while but what's been holding me down is that the official spec for the theme indicates that variables have to be the same color as text. However, a good idea would be to add a custom variable in the theme for enabling this for people that prefer it.

sulami:
Honstly, I'm fine with anything, I've also noticed that the anchors have different colours between the themes, which apparently is a `function-name`, but I'll leave the decision on how resolve this to Henrik I guess. I just thought sending in a PR is better (and probably faster) than just complaining that the colours are different (and worse imho) in the dark version.

ema2159:
I get it. Though I'm the one with the responsibility to maintain this theme, so I can help you out with the PR. You could do take a similar approach to the one taken with the `doom-solarized-dark-brighter-text` variable! Check it out right there in the same file.

hlissner:
I'd prefer that variables have a different color than the foreground text. I don't mind if we deviant from the spec in this regard. This is what I'm doing in `doom-one` and `doom-dracula` and, for consistency, would like to extend it to other themes.

I don't think it necessitates another variable though. Themes are already very customizable with the likes of `custom-set-faces` and `set-face-attributes`, and work fine for this simple changes.

ema2159:
@hlissner understood! It's fine then.

PR #555, opened 2021-01-02
Solarized dark: customizable color of constants

lukasz-m-maciejewski:
Adding a customization option for solarized dark theme that allows one to make the color of the constants blue instead of magenta. The reason for it is that I work with C++ alot and in C++ code, a lot of things (primary example: namespace names) are considered constants. This results in a lot of magenta on screen, which is a bit hard on the eyes.

### DIFF
--- a/themes/doom-solarized-dark-theme.el
+++ b/themes/doom-solarized-dark-theme.el
@@ -21,6 +21,11 @@
   :group 'doom-solarized-dark-theme
   :type 'boolean)
 
+(defcustom doom-solarized-dark-make-constants-blue nil
+  "If not-nil, makes constants blue instead of magenta"
+  :group 'doom-solarized-dark-theme
+  :type 'boolean)
+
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the suggestion but I cannot accept it. Requests to fix syntax highlighting in a specific mode are too niche and can have an unpredictable effect on other major modes that use the same faces. I also want to avoid adding variables for every little change. I'm working on a system so folks can change these color variables, but until then, use `custom-set-faces` to customize the theme. e.g.

```lisp
(load-theme 'doom-solarized-dark t)

;; must be used after the theme as loaded
(custom-set-faces `(font-lock-constant-face (( t (:foreground ,(doom-color 'blue))))))
```

If you use Doom, load order isn't important if you use the `custom-set-faces!` macro:

```lisp
(custom-set-faces! `(font-lock-constant-face :foreground ,(doom-color 'blue)))
```

Hope that helps!

PR #558, opened 2021-01-06
Update plain-theme and add plain-dark-theme

das-s:
# Description

Motivation for this change is supplying a dark theme for plain-theme as well as improving the current plain theme.

This PR adds support for plain-dark-theme which is the sister theme complement to the existing plain theme. Use of italics according to the [original plain theme](https://github.com/gko/plain/) has been adapted, especially for org-mode headers. Italics fit into the minimalistic theme approach better than the current bold ligatures as they provide a rather subtle way of emphasis beside colors which this theme naturally lack.  

# Screenshots
<details><summary>Code</summary>
<p>

### before

</details>
</p>

<details><summary>Org</summary>
<p>

### before

</details>
</p>

### DIFF
--- a/themes/doom-plain-dark-theme.el
+++ b/themes/doom-plain-dark-theme.el
@@ -1,0 +1,170 @@
+;;; doom-plain-dark-theme.el --- inspired by gko's plain theme for VSCode
+
+(require 'doom-themes)
+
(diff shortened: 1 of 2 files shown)


ema2159:
Nice! Thank you very much!

PR #562, opened 2021-01-27
Fix doom-zenburn.el typo

drfiresign:
`doom-zenburn.el` file left out a single quote on two `:weight` attributes in the helm section. This fixes the typo and restores the theme's ability to be loaded and unloaded. Without the fix users cannot load new themes after loading doom-zenburn.


ema2159:
Perfect. Thanks!

PR #576, opened 2021-02-28
Fix doom-horizon's blue background in terminal and modeline fg bug.

### LINKED ISSUE #438: Horizon theme looks broken in terminals
Hi!
I wanted to set `doom-horizon` as my main theme but found that it uses weird bg color in terminals:

20200411_06h56m48s_grim

I've tried to fix it by setting bg color to nil in the theme but looks like it's not the right way to do it since it also messes with other colors as well.

Some details: emacs 26.3, term kitty (same issue also in alacrity and iterm2).

daanturo:
This PR fixes #438.

Before the fix, without and with `-debug-init`:

Screenshot from 2021-02-28 16-24-01
Screenshot from 2021-02-28 16-24-13

After fixing:

Screenshot from 2021-02-28 16-24-59


hlissner (changes requested):
Could you fix the modeline too? A bright white in a dark theme is hardly appealing.

daanturo:
I also changes some more colors to avoid that dazzling blue light in the terminal.

Screenshot from 2021-03-01 13-05-02

Screenshot from 2021-03-01 13-02-06

daanturo:
Ping @hlissner ! I have completed you request, may you consider again?

hlissner:
Sorry for the late response. Thanks for your help!

PR #580, opened 2021-03-18
gruvbox-light: lsp symbol highlight bold, no background

practicalli-john:
lsp enables symbol highlighting by default.  In the doom-gruvbox-light theme, the highlight color makes the text harder to read as there is insufficient contrast.  

In the doom-gruvbox-theme the `lsp-face-highlight-read` defines a background color as the highlight, using `(doom-blend bg orange 0.5)`.  
The `lsp-face-highlight-textual` inherits from `lsp-face-highlight-read` and is used to highlight symbols by lsp

The change removes the background color from `lsp-face-highlight-read` and adds a bold weight to that face

Original theme
doom-gruvbox-light-orange

Proposed change
doom-gruvbox-light-bold

### DIFF
--- a/themes/doom-gruvbox-light-theme.el
+++ b/themes/doom-gruvbox-light-theme.el
@@ -447,7 +447,7 @@ background contrast. All other values default to \"medium\"."
 
    ;; lsp
    (lsp-ui-doc-background      :background base2)
-   (lsp-face-highlight-read    :background (doom-blend bg orange 0.5))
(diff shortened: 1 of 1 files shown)


hlissner:
I agree the old background was too harsh, but I think bolding alone is difficult to see. The contrast has been significantly reduced in d4ce760, which should be easier on the eyes without harming legibility. Thanks for bringing it to my attention!

Issue #572, opened 2021-02-21
doom-dracula requires reload for proper rendering


prashantvithani:
`doom-dracula` theme renders different colors for certain text after `reload-theme (SPC h r t)`

#### before reload
![dracula-before-reload](https://user-images.githubusercontent.com/5552382/108622040-506bfe00-745c-11eb-9f5f-b2f19532d375.png)

### after reload
![dracula-after-reload](https://user-images.githubusercontent.com/5552382/108622049-58c43900-745c-11eb-9368-5fd7b5065a83.png)

## Steps to reproduce
1. Start doom with `doom-dracula`
2. Paste following snippet in scratch with `ruby-mode`
   ```ruby
	  # @param param1 [String]
	  # @param param2 [Integer]
	  def method(param1, param2)
	    @param1 = param1
	    @param2 = param2
	    param3 = "#{@param1}_#{@param2}"
	    @param3 = param3
	    @param4 = {
	      key1: @param1,
	      key2: "hello",
	      key3: 123,
	    }
	    if @param[:key1].is_a?(Hash)
	      true
	    else
	      false
	    end
	  end
   ```
2. reload the theme `(SPC h r t)`

I had expected the theme colors rendered *after reload* to have rendered by default. I'm wondering what's going wrong. Even calling `(doom/reload-theme)` in config.el won't work.

## System Information
+ Emacs version: 28.0.50 
+ Emacs Distro: Doom Emacs 
+ Operating System: Arch 
+ Installed commit of doom-themes: 47925933

sebastianappler:
Thank you for pointing this out! 

I have the same problem. Turns out its a problem with dracula theme, all text is white in yaml files before running `reload-theme (SPC h r t)`.

PR #584, opened 2021-04-03
dracula-theme: Fix rendering of variable color

sebastianappler:
The value passed to `doom-lighten` function was incorrectly passed as variable reference `'magenta` instead of `magenta`.

This caused the color for `variables` to not render during startup but for some reason it worked if you did a `reload-theme (SPC h r t)`. See issue #572.

I have tested that this works in my fork.


hlissner (approved):

sebastianappler:
@hlissner can we merge or should I do anything?

PR #585, opened 2021-04-07
doom-horizon: make line-number-current-line closer to the original

daanturo:
Make it closer to the original theme and improve aesthetic.

On the left is Horizon on Visual Studio Code, the upper right is on Emacs, before, the lower one is also Emacs, but after brightening.
(Notice current line on each).

image


hlissner:
Thanks!

PR #596, opened 2021-05-05
Change function color to cyan (88C0D0, nord8) according to official doc.

ianpan870102:
Nord's official documentation (https://www.nordtheme.com/docs/colors-and-palettes) states that nord8 `#88C0D0` should be used for functions and methods. Currently, the color nord7 `#8FBCBB` is used. I submitted a PR to fix this just now.

Screenshots below. The changes are subtle.

## Before

image

## After

image


ianpan870102:
Cheers!

PR #601, opened 2021-05-23
Add Monokai Pro Machine, Ristretto, Octagon, Spectrum

minikN:
Hello,

I've added the missing filters:
- Octagon
- Ristretto
- Machine
- Spectrum

# Spectrum
image

# Ristretto
image

# Machine
image

# Octagon
image

PR #532, opened 2020-10-25
Add missing filters; normalize colors

minikN:
Hello,

I've added the missing filters:
- Octagon
- Ristretto
- Machine
- Classic (Although there already is a classic variant, this one is slightly different. It's Hazenbergs own variant I guess
- Pro (I've converted the pro theme to match the other filters. The colors where not quite correct and it will benefit from the configuration I've already done)

I've also normalized colors throughout the filters. Unfortunately the original Palette doesn't provide all colors we need for the doom theme so I need to arbitrarily pick colors in between. I've gone over my previous PR again and unified what colors I pick for each filter so they don't appear random.
To put it simple: Every color for every filter I needed to add now has the same "distance" from the colors provided by the original palette itself making the filters look more uniform.
- Pro
image

- Pro (Classic)
image

- Spectrum
image

- Ristretto
image

- Machine
image

- Octagon
image

PR #602, opened 2021-05-23
Add Monokai Pro and Classic 

minikN:
Hello,

I've added the missing filters:
- Classic (Although there already is a classic variant, this one is slightly different. It's Hazenbergs own variant I guess
- Pro (I've converted the pro theme to match the other filters. The colors where not quite correct and it will benefit from the configuration I've already done)

# Pro
image

# Pro (Classic)
image


minikN:
@hlissner  @ema2159  @ianpan870102  Please check if everything is like you wanted it in this PR and #601

minikN:
Reference: #532

ema2159:
Sure! The one thing I don't like much is the change in the bg-alt. The contrast is too much, it's too dark. The same would go for the other filters' implementation.

minikN:
> Sure! The one thing I don't like much is the change in the bg-alt. The contrast is too much, it's too dark. The same would go for the other filters' implementation.

I'll review it. From the top of my head I can't tell you where I picked the color from. Might have been implemented like this in VS Code (where I took the color from) or I chose it myself because it was missing.

If the latter is the case, I'll adjust it. But this will take a couple of days as I'm super stumped with work atm.
Alternatively, your or @ianpan870102 can recommend a different color.

ema2159:
I think I can pick some from VS Code. I think I know which ones you picked, but there is another option to pick also in VS Code which I'd rather choose.

minikN:
bump

hlissner (changes requested):
Sorry for the late response. Would you mind resolving the conflicts? Then I think this will be ready to merge.

minikN:
Sorry but I don't use doom/doom-themes anymore.

PR #608, opened 2021-05-31
doom-one-light: use correct symbol name

dalugm:
fix `Symbol’s value as variable is void: doom-modeline-brighter-modeline`


hlissner (approved):

hlissner:
Good catch. Thanks!

PR #609, opened 2021-05-31
molokai & monokai-classic : set vc-modified to cyan after 3f09923

aagon:
Very simple change, did not follow the template for that reason.

3f09923 forced git-gutter faces to follow vc-modified, but in a few themes, vc-modified was (mistakenly I guess) set to base4 (a kind of black), hence I am modifying vc-modified to cyan.

Feel free to ask any question.

Best,

Aymeric


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #614, opened 2021-06-02
Add My Doom Gruvbox

ItsNilDev:
> Please follow these steps before submitting your PR
>
> - [x] Include before and after screenshots if your change is visual: 
Before: 
before
After: 
after

> - [x] Please explain what the PR does and why:
As you can see, it makes it to look better for the eyes and feel great!
The code is a bit ugly but I hope we can make it better!

PR #616, opened 2021-06-03
Add My Doom Gruvbox (fix)

ItsNilDev:
Ok, after my previous PR (#614) , I realized that my doom-themes was out-of-date, and everything was looking broken so I decided to make my theme again from 0. The images look the same ( the look of my changes didn't change ), so it still seems like the images in #614.
I decided to break down my changes in  5 parts so you can I understand better what I changed: 

- **Added some colors**, I added 2 more colors based on morhetz's gruvbox theme (vim) which are a blue and black.
-  **Changed face categories**, Actually I changed functions, methods, operators, and type. But I changed them because they looked cool, it's not necessary to keep these.
- **Changed modeline's face**, In original doom-gruvbox the modeline looks like this:
image
It's actually hard to recognize in which window you are + the modeline doesn't look well with other colors, so I changed it and now it looks like:
image
- **Changed second depth of Rainbow delimiters face**, Currently rainbow delimiters's second and third faces are somehow not recognizable from each other as you can see here: 
image
So I changed it and now it looks like this:
image
- **And Finally I changed which-key faces**, right now, which-key faces are not obvious, so I added some colors to it.
Before: 
image
After:
image

### DIFF
--- a/themes/doom-gruvbox-theme.el
+++ b/themes/doom-gruvbox-theme.el
@@ -59,7 +59,6 @@ background contrast. All other values default to \"medium\"."
    (magenta     '("#cc241d" "#cc241d" "magenta"))       ; red
    (violet      '("#d3869b" "#d3869b" "brightmagenta")) ; bright-purple
    (orange      '("#fe8019" "#fd971f" "orange"))        ; bright-orange
-   (my_orange   '("#d65d0e" "#d65d0e" "orange"))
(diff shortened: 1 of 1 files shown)


ItsNilDev:
Should be valid..

hlissner (approved):

hlissner:
Thanks for your help!

PR #622, opened 2021-06-14
doom-horizon: Differentiate "blue" from "teal"

daanturo:
Before: above, after: below

image

Previously, `doom-horizon` defined `blue` with exactly the same color code as `teal`, which led to some visual confusions when `teal` and `blue` are placed next to each other.

For example, in `doom-themes-base.el`:
```lisp
    (rainbow-delimiters-depth-5-face :foreground teal)
    (rainbow-delimiters-depth-6-face :foreground blue)
```
Then parentheses of level 5 and 6, when colorized with [`rainbow-delimiters`](https://github.com/Fanael/rainbow-delimiters) are indistinguishable.

This PR attempts to make `blue` standout from `teal` a bit, enough to tell the two a part by assigning `SkyBlue` to `blue`.


ema2159:
I would change the teal variable instead of the blue one since usually what's done is, when the teal color is missing (which is more often than not) any other color is duplicated into it (either cyan, blue, etc). So I would change the teal instead of the blue.

daanturo:
> when the teal color is missing (which is more often than not) any other color is duplicated into it (either cyan, blue, etc)

Can you elaborate more about this, I don't thoroughly understand it.

Did you intent to change 'teal to another (real-world) shade of blue? Since out the those colors, ~1/3 of them are just "cyan"s.
image

Anyway, other users and I would appreciate any forms of visual changes as you think which needs to.

ema2159:
Yes. The thing is that teal is always an extra color used to solve contrast issues for some themes and a lot of times it just has the same color as some other color (blue, cyan, etc), whereas the blue color is actually a color from the theme's palette which is considered in the design of the theme. So, if one color should be changed, it is teal rather than blue.

daanturo:
> if one color should be changed, it is teal rather than blue.

image

This is the new difference, can you review again?

PR #628, opened 2021-06-27
Add vertico-current definition

bdarcus:
This adds support for [vertico](https://github.com/minad/vertico); specifically highlighting the current selected candidate via the `vertico-current` face.

I simply copy-and-pasted the definitions for `selectrum`, which `vertico` is basically a drop-in replacement for.

The result, therefore, as with `selectrum`, is a more subtle highlighting:

Screenshot from 2021-06-27 17-27-35

Before:

Screenshot from 2021-06-27 17-30-27

### DIFF
--- a/doom-themes-base.el
+++ b/doom-themes-base.el
@@ -789,6 +789,8 @@
     (ivy-posframe-border :inherit 'internal-border)
     ;;;; selectrum
     (selectrum-current-candidate :background region :distant-foreground nil :extend t)
+    ;;;; vertico
+    (vertico-current :background region :distant-foreground nil :extend t)
     ;;;; jabber
     (jabber-activity-face          :foreground red   :weight 'bold)
     (jabber-activity-personal-face :foreground blue  :weight 'bold)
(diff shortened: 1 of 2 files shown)


iyefrat:
@ema2159 Do you know when you'll have time to look this over and merge it? It's a bit of a blocker for some stuff in the selectrum pr (hlissner/doom-emacs#4664).

minad:
@iyefrat Why do you consider this to be a blocker? This is only eye-candy, which does not affect the functionality at all, isn't it?

ema2159:
I think I can do it just now! Usually, these PRs don't give much trouble.

bdarcus:
> Why do you consider this to be a blocker? This is only eye-candy, which does not affect the functionality at all, or does it?

@minad - just a guess: maybe not a traditional blocker, but he wants a clean and consistent UI before merging an experimental vertico flag? 

Also, I didn't test, but from what he said, the previous behavior may be more problematic in other themes (the screenshot just shows `doom-one`)?

iyefrat:
@minad The color used for the first orderless matching face in the theme I'm using (`doom-city-lights`!) is too similar to the blue used for highlighting, making partially matched candidates hard to read. I couldn't figure out how to change this outside of the theme definition due to odd errors around the fact that `region` is a color here, but isn't acknowledged as one elsewhere. It did make me realize that I should probably look into tying the orderless matching colors to the theme though.

PR #632, opened 2021-07-07
doom-manilla: add theme

obar:
A variant of Solarized Light with more contrast. A lighter background, darker foreground, and some color tweaks.


obar:
Is there a setup function to make [an official screenshot](https://github.com/hlissner/emacs-doom-themes/tree/screenshots)? At any rate here are a quick pair to show what the theme looks like.

Screenshot from 2021-07-07 17-12-45

Screenshot from 2021-07-07 17-11-47

hlissner:
Thanks for the submission, but we are not currently accepting new themes as our API is about to undergo a total rewrite. Side note: your theme doesn't adhere to the formatting of our solarized themes (i.e. you used an old version of our themes as a basis for yours) -- in particular, package faces are not properly commented and comments were not carried over (Granted, this formatted was only introduced last month, so it's understandable that you missed it).

Sorry for the trouble!

obar:
Ah I see. I've been using this for almost a year now (making small tweaks along the way) so I branched things off a while ago. It's not quite a `solarized-light-high-contrast` theme because I've changed the colors too much. I thought I would finally share what I've got but that definitely makes sense about not adding new themes! Thanks for taking a look at this and sorry for the inadvertent line noise ;)

PR #639, opened 2021-07-26
Add marginalia file privilege faces

tecosaur:
Now that https://github.com/minad/marginalia/pull/91 has been merged, Marginalia has new faces to do exactly the same thing a `diredfl` currently does with file attributes. I.e. this:

image

So, here's a PR that just copies the `diredfl` face information, just for the incoming Marginalia faces.

### DIFF
--- a/doom-themes-base.el
+++ b/doom-themes-base.el
@@ -932,6 +932,14 @@
     (markdown-html-entity-face        :inherit 'font-lock-variable-name-face)
     (markdown-html-tag-delimiter-face :inherit 'markdown-markup-face)
     (markdown-html-tag-name-face      :inherit 'font-lock-keyword-face)
+    ;;;; marginalia
+    (marginalia-file-priv-dir   :foreground blue)
+    (marginalia-file-priv-exec  :foreground green)
+    (marginalia-file-priv-link  :foreground violet)
+    (marginalia-file-priv-other :foreground magenta)
+    (marginalia-file-priv-rare  :foreground fg)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

protesilaos:
Hello @tecosaur!  This relates to what we discussed earlier at the marginalia repo: https://github.com/minad/marginalia/pull/91#issuecomment-886821035

I think there is one small change that could be done which would (i) not change the character of the theme, and (ii) ensure greater consistency between diredfl and marginalia.  Please note that I am not familiar with the theme and thus cannot suggest more thoroughgoing reforms.

The diff (in case you want to send a patch):

```diff
@@ -439,7 +439,7 @@
     (diredfl-ignored-file-name      :foreground comments)
     (diredfl-link-priv              :foreground violet)
     (diredfl-no-priv                :foreground fg)
-    (diredfl-number                 :foreground orange)
+    (diredfl-number                 :foreground violet)
     (diredfl-other-priv             :foreground magenta)
     (diredfl-rare-priv              :foreground fg)
     (diredfl-read-priv              :foreground yellow)
```

Before VS After:

Screenshot from 2021-07-26 22-09-31

Screenshot from 2021-07-26 22-19-01

I would not do it the other way around (add orange to `marginalia-size` or `marginalia-number`) because I think that draws more attention to the number than what it really needs.

tecosaur:
Thanks @protesilaos! I'll have a look at that :slightly_smiling_face:

PR #640, opened 2021-07-26
Add orderless faces

tecosaur:
This adds support for the four (orderless)[https://github.com/oantolin/orderless] faces, seen below

image

Unfortunately, this patch makes my Emacs segfault, so I can't easily provide an after image. If anyone has any idea what's going on *please* let me know. Full error: http://ix.io/3u7E

### DIFF
--- a/doom-themes-base.el
+++ b/doom-themes-base.el
@@ -1044,6 +1044,11 @@
     ;;;; objed
     (objed-mode-line :inherit 'warning :weight 'bold)
     (objed-hl        :inherit 'region :background (doom-blend region bg 0.5))
+    ;;;; orderless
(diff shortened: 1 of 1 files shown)


tecosaur:
Oh you've got to be kidding me. The segfaulting was caused by me doing `'green` instead of `green` etc.

tecosaur:
It's fine now, here's an after screenshot:
image

tecosaur:
Maybe a different blue would be better?

tecosaur:
I've decided to try something [modus inspired](https://raw.githubusercontent.com/minad/corfu/main/screenshot.png)

image

hlissner (approved):

PR #641, opened 2021-07-26
Add marginalia documentation face

tecosaur:
This defaults to `:inherit completions-annotations`, which is `(italic shadow)`. However, for *documentation* `font-lock-doc-face` would seem more sensible.

Default:

image

With this patch:

image


tecosaur:
NB: This was discussed with minad and he said he'd keep it as is with Marginalia, and leave it up to themes to change the documentation face.

hlissner:
Makes sense to me. Thanks!

Issue #659, opened 2021-09-15
Ayu Light Line Number Invisible


CramMK:
When using the doom-ayu-light color theme, everything works except the line numbers (see screenshot).
I presume that the color `base4` just happens to be the same as the background and some how no one noticed it.

![image](https://user-images.githubusercontent.com/39596050/133397094-67c72edc-7527-402f-b9e2-29d5cb5e3b60.png)

I propose following upstream diff:
```
diff --git a/themes/doom-ayu-light-theme.el b/themes/doom-ayu-light-theme.el
index 4a7f8bf..b0e21ff 100644
--- a/themes/doom-ayu-light-theme.el
+++ b/themes/doom-ayu-light-theme.el
@@ -141,7 +141,7 @@ determine the exact padding."
    (modeline-bg-inactive-l `(,(car bg) ,@(cdr base1))))

   ;;;; Base theme face overrides
-  (((line-number &override) :foreground base4)
+  (((line-number &override) :foreground (doom-lighten fg 0.7))
    ((line-number-current-line &override) :foreground fg)
    ((font-lock-comment-face &override)
     :background (if doom-ayu-light-comment-bg (doom-lighten bg 0.05)))
```
Which looks like this:

![image](https://user-images.githubusercontent.com/39596050/133397562-355f5fe3-d0bc-4b77-b9e0-77e1da910963.png)

## System Information
+ Emacs version: 28 and 27.2
+ Operating system: Fedora 34 and 33

CramMK:
Fixed with #638

PR #660, opened 2021-09-15
Fix ayu-light line-number

CramMK:
As described in #659 (Screenshots also there):

Ayu-Light Variant has invisible line numbers, for some reason.
Just use a lighter variant of the foreground to fix that.


iyefrat:
The other ayu-themes PR I just merged includes a fix for this, albeit with a different color. Can you unpin `doom-themes` and give it a spin?

CramMK:
Oh my bad then, i only looked through the issues for that and overlooked that PR. I'll close both the PR and issue.
Thank you!

PR #681, opened 2021-11-10
Make window splits more visible for oceanic-next-theme

ngharrison:
This small change makes divisions between windows more visible.

Before:
image

After:
image

That's it!

### DIFF
--- a/themes/doom-oceanic-next-theme.el
+++ b/themes/doom-oceanic-next-theme.el
@@ -61,7 +61,7 @@ determine the exact padding."
 
    ;; face categories -- required for all themes
    (highlight      yellow)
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must turn it down because the darker window divider is intentional. Perhaps I'll make it darker to increase it's contrast, but I don't want to make it lighter.

To change it in your private config:

```elisp
(with-eval-after-load 'doom-themes
  (custom-set-faces `(vertical-border ((t (:foreground ,(doom-darken 'base4 0.25)))))))
```

Or if you use Doom Emacs:

```elisp
(custom-theme-set-faces! 'doom-oceanic-next `(vertical-border :foreground ,(doom-darken 'base4 0.25)))

PR #693, opened 2021-12-21
docs: clean up theme table

jeetelongname:
I noticed that the theme table was slightly out of order and earl-grey
was missing.

I have fixed the former and added in the latter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,15 +81,17 @@ maintain them and address inconsistencies:
 | Name                                | Description                                                                                                  |
 |-------------------------------------|--------------------------------------------------------------------------------------------------------------|
 | `doom-1337`                         | ported from [VSCode's 1337 Theme][vscode-1337]                                                               |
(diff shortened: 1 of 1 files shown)


iyefrat:
Don't bother resolving the conflict yet, I need to fix `doom-Iosvkem` before I can merge this. I'll ping you once I've done that.

jeetelongname:
k thanks, (love ya)

hlissner:
> I need to fix doom-Iosvkem before I can merge this

@iyefrat What fix is this referring to?

jeetelongname:
doom-Iosvkem is the only theme that is capitalised, its annoying because that puts the table out of wack in terms of order and it should not be like that anyway, But @iyefrat wanted an elegant way to deprecate that name while also decapitating it

hlissner:
I went ahead and did it in 142554f. Thanks for reminding me!

PR #695, opened 2021-12-26
Fix company's scrollbar in solarized-light theme

cordelia04m:
The old one has "wheat" and "red" colors, which look out of place with the theme.

Before
Screenshot_20211226_162303

After
Screenshot_20211226_162219


cordelia04m:
Hold on. It looks like doom-one-light has that wheat and red colors as well. I guess I'll add the change to the base theme instead.

PR #700, opened 2022-01-05
Fix illegible org blocks: doom-tokyo-night

brongulus:
## Before:
tokyo-before

## After:
tokyo-after

Fixes the illegible colour of org-blocks that stands out and causes issues when reading commented/greyed out characters.


hlissner (approved):

hlissner:
Thanks for your help!

PR #703, opened 2022-01-27
theme: add doom-nord-aurora

MoskitoHero:
I love the Nord theme, but I think it is a bit too monochrome. So I decided to use the [Nord Aurora palette](https://www.nordtheme.com) and create my own. I have been using it for about a year now, and it is very usable, so I think it is time to share it. 

Screenshot_20220127_071925


hlissner (approved):

hlissner:
Sorry for the late response. Thanks for the submission!

PR #705, opened 2022-02-14
fix(ext-visual-bell): Make it work with emacs 29

tsdh:
Since emacs 29 (the current git master), the `mode-line` face is the parent of the new face `mode-line-active` and `mode-line-inactive`.  For remapping purposes, the `mode-line-active` face has to be used, see details at:

http://debbugs.gnu.org/cgi/bugreport.cgi?bug=53636

I've also took the leisure to add a closure to the timer instead of providing arguments to the timer function since the package requires emacs 25.1 anyhow, so support for `lexical-binding` can be assumed.


hlissner:

PR #713, opened 2022-03-10
fix(gruvbox-light) make gray readable

isker:
This makes gray readable by way of overriding vterm colors to match the
kitty terminal's gruvbox-light theme
https://github.com/kovidgoyal/kitty-themes/blob/6b06ed8661c62b1fcc63c9de9054d9cc12c319f6/themes/gruvbox-light.conf,
which itself is using colors from the upstream gruvbox palette
https://camo.githubusercontent.com/d080d9c204408ef06b862b76bc795f930b3a9b1be4c5d2de149f1d8eb765b660/687474703a2f2f692e696d6775722e636f6d2f3439714b7959572e706e67.

Unfortunately the variable names associated with this palette are all
over the place in this theme.

Before:
before

After:
after

(I have no idea why fish's set_color prints them in this order...)

### DIFF
--- a/doom-themes-base.el
+++ b/doom-themes-base.el
@@ -1374,8 +1374,6 @@
     ;;;; volatile-highlights
     (vhl/default-face :background grey)
     ;;;; vterm
-    (vterm               :foreground fg)
-    (vterm-color-default :foreground fg)
     (vterm-color-black   :background (doom-lighten base0 0.25)   :foreground base0)
     (vterm-color-red     :background (doom-lighten red 0.25)     :foreground red)
     (vterm-color-green   :background (doom-lighten green 0.25)   :foreground green)
--- a/themes/doom-gruvbox-light-theme.el
+++ b/themes/doom-gruvbox-light-theme.el
@@ -474,8 +474,6 @@ background contrast. All other values default to \"medium\"."
     :background modeline-bg-inactive-l
     :box (if -modeline-pad `(:line-width ,-modeline-pad :color ,modeline-bg-inactive-l)))
    ;;;; vterm
-   (vterm               :foreground fg)
-   (vterm-color-default :foreground fg)
    (vterm-color-black   :background grey    :foreground base1)
    (vterm-color-red     :background red     :foreground faded-red)
    (vterm-color-green   :background green   :foreground faded-green)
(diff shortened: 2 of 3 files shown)


isker:
Somewhat related:  why are the `term-` colors defined differently than the `vterm-` colors in the base?  Don't these do the same things in different modes?

https://github.com/doomemacs/themes/blob/83fd9545c2823b4b2610947802fa3a52995517b4/doom-themes-base.el#L1322-L1332
https://github.com/doomemacs/themes/blob/83fd9545c2823b4b2610947802fa3a52995517b4/doom-themes-base.el#L1381-L1391

Should one be redefined in terms of the other?

isker:
Somewhat-somewhat related:  what does `vterm-color-default` do?  I can't find a usage of it anywhere.

hlissner (changes requested):
> Somewhat related: why are the term- colors defined differently than the vterm- colors in the base? Don't these do the same things in different modes?

These faces have changed a lot in the past year and in the 28.1 release. Now that vterm-color-* all inherit from term-color-*, which inherit from ansi-color-*, I think we can get rid of all our defaults for term/vterm in the base theme, and leave only our ansi-color-* faces.

> what does vterm-color-default do? I can't find a usage of it anywhere.

It's been some time since I checked, but I believe that controls the fallback background/foreground. Then came `vterm`. Seems both were removed in any case, so go ahead and remove them.

isker:
> These faces have changed a lot in the past year and in the 28.1 release. Now that vterm-color-* all inherit from term-color-, which inherit from ansi-color-, I think we can get rid of all our defaults for term/vterm in the base theme, and leave only our ansi-color-* faces.

I see now the item about ANSI color faces in emacs 28.1's NEWS, but I don't see any such faces used in this repo.  Are you asking me to do something here?

> It's been some time since I checked, but I believe that controls the fallback background/foreground. Then came vterm. Seems both were removed in any case, so go ahead and remove them.

Will do.  Here's where `vterm-color-default` went away at least:  https://github.com/akermu/emacs-libvterm/commit/c15c1a5615420defd95e000829b658b09509ced1

hlissner (approved):

hlissner:
Sorry for the late response. Thanks for looking into this and for your help!

> I see now the item about ANSI color faces in emacs 28.1's NEWS, but I don't see any such ansi-color faces used in this repo. Are you asking me to do something here?

My bad, I think I was talking to myself at you! I wasn't asking you to do anything.

I was musing about dropping our `vterm-color-*` and `term-color-*` face definitions, and replacing them with `ansi-color-*` faces (which they both inherit form), but that's something I'll look into later, once I figure out how to do so and remain compatible with all the Emacs versions in between 25.1 and 28.1.

Anyhow, thanks again!

PR #716, opened 2022-03-24
org-mode: align list items in mixed-pitch-mode

psii:
This pull request fixes the alignment of list items in org-mode when using `mixed-pitch-mode` or similar modes like doom's zen mode. It does so by giving the space characters before the list bullet item and the single space after the bullet item the `org-indent` face.

Org buffer with standard fixed-pitch font:
scr-b

Org buffer with doom's zen mode activated before the patch:
scr-a

Org buffer with doom's zen mode activated after the patch:
scr-c


hlissner (approved):

hlissner (approved):

hlissner:
Makes sense to me. Thanks for your help!

PR #775, opened 2023-02-19
feat(base): add embark-target face

nbarrientos:
Embark [remaps](https://github.com/oantolin/embark/blob/1f6c87d79f5d138cd128c230d727c796b4bc46cd/embark.el#L3333) `vertico-current` to `embark-target` [0] so to make the selected candidate highlighting consistent it makes sense to align how Doom themes customises `embark-target`, too.


doom-themes-base.el
```diff
@@ -816,6 +816,8 @@
     (ivy-posframe-border :inherit 'internal-border)
     ;;;; selectrum
     (selectrum-current-candidate :background region :extend t)
+    ;;;; embark
+    (embark-target :background region :extend t)
```
hlissner, line 820: Could you move this up above `;;;; enh-ruby-mode ...` and below `;;;; elscreen` (i.e. [above this line](https://github.com/doomemacs/themes/blob/b5ff201f4bea4286e9ed015a2043cf2394182232/doom-themes-base.el#L516))? These sections are ordered alphabetically.

hlissner: Also, use hard inheritance here:

```lisp
((embark-target &inherit vertico-current))
```

The difference between this and `:inherit` is that `&inherit` copies the inherited face's attributes when the theme is compiled, rather than loaded, and allows it to work even if vertico is never loaded/installed.

hlissner (changes requested):

doom-themes-base.el
nbarrientos: Thanks for the feedback! I've incorporated all the suggested changes.

hlissner:
LGTM! Thanks for your help!

Issue #753, opened 2022-09-07
Warnings: nil value is invalid


seagle0128:
### Theme(s)

All themes (base)

### What were you expecting?

No warnings

### What actually happened?

In recent 29.1, these warnings come out.

```
Warning: setting attribute ‘:background’ of face ‘ivy-minibuffer-match-face-1’: nil value is invalid, use ‘unspecified’ instead.
...
```

### Steps to reproduce

1. Installed doom-themes.
2. Load doom-one theme.

### Operating system

All

### Emacs version

29.0.50

### Installed commit of doom-themes

latest

elken:
Think this is related to https://github.com/emacs-mirror/emacs/commit/d0a66f3

AFAIU I don't think we're planning to support HEAD until things start to stabilize a bit.

seagle0128:
Make sense...

hlissner:
This issue should now be resolved. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

Ethanlinyf:
I also still have the same issue: 

Warning: setting attribute ‘:background’ of face ‘font-lock-comment-face’: nil value is invalid, use ‘unspecified’ instead.

emacs-plus@29 on MacBook Pro 2021 with macOS, Ventura 13.1

seagle0128:
I think this may cause the issue. e.g.

```elisp
((font-lock-comment-face &override) :background (if doom-feather-light-brighter-comments base0) :italic t)
```

BobenHome:
> Hi @hlissner I still got this error:
> 
> ```
> Warning: setting attribute ‘:background’ of face ‘font-lock-comment-face’: nil value is invalid, use ‘unspecified’ instead.
> ```

How did you fix it later?

hlissner:
I've slowly been stomping out the lingering `nil`'s in our themes. Though there may be a couple more, I *think* I can consider this issue mostly resolved for now. Feel free to follow up here if you find an issue that still emits this warning, however.

PR #777, opened 2023-03-09
fix: handle all nil => unspecified cases

elken:
A recent change to Emacs 29 makes setting a face attribute to nil now
throw a warning, so using some advice we recur through a face spec (only
when there's a valid spec) and change all the nils to 'unspecified.

I'm aware this probably isn't the _best_ way to handle it, I'm open to suggestions here.

Fixes: #753


hlissner:
`doom-themes--custom-set-unspecified` was a step further than I was willing to go to address this (I'd rather swap out all the `nil`'s in the themes), but I've absorbed the other two changes into d07c54e. Thanks for your help, in any case!

PR #818, opened 2024-01-09
Fix: show-paren fg val from 'nil' to 'unspecified'

yaaama:
This commit fixes a warning Emacs sends regarding the foreground value being set to nil for show-paren faces in the gruvbox theme.

The faces that cause the warning are: 'show-paren-match' and 'show-paren-mismatch'.
Instead of setting it to nil, it is now 'unspecified'.

Common issue it seems like with the new changes to Emacs.

-----
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [ ] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).


hlissner (approved):

hlissner:

Issue #827, opened 2024-05-05
Add tokyo-night storm and moon


drishal:
### Describe your request

as the title says, if possible please add tokyo night storm and moon variants as well 

### Resources

https://github.com/folke/tokyonight.nvim

Luxvao:
+1

i-am-linja:
Can we please also add Day? (Or Light as it's known in VSCode.) Light mode users do exist.

foster-hangdaan:
I can work on this if possible. I can add the Storm and Day variants. However, I probably won't do the Moon variant as it doesn't seem to be a variant that is listed on the [official repo](https://github.com/tokyo-night/tokyo-night-vscode-theme).

azzamsa:
@foster-hangdaan The `moon` variant is the most popular, as it is used by the author of `folke/tokyonight.nvim` himself.

While `https://github.com/tokyo-night/tokyo-night-vscode-theme` (enkia) was the original, you'll notice that `https://github.com/folke/tokyonight.nvim` is more widely adopted, and often serves as a reference for other implementations.

- `https://github.com/folke/tokyonight.nvim`: 6.1k stars
- `https://github.com/tokyo-night/tokyo-night-vscode-theme`: 1.7k stars

azzamsa:
For any `folke/tokyonight` moon variant users, before it gets ported to Doom, you can use `doom-moonlight`, which is similar.

foster-hangdaan:
> @foster-hangdaan The `moon` variant is the most popular, as it is used by the author of `folke/tokyonight.nvim` himself.
> 
> While `https://github.com/tokyo-night/tokyo-night-vscode-theme` (enkia) was the original, you'll notice that `https://github.com/folke/tokyonight.nvim` is more widely adopted, and often serves as a reference for other implementations.
> 
>     * `https://github.com/folke/tokyonight.nvim`: 6.1k stars
> 
>     * `https://github.com/tokyo-night/tokyo-night-vscode-theme`: 1.7k stars

I'll add the Moon variant to #835 when time permits in order to close this issue in one shot.

drishal:
> > @foster-hangdaan The `moon` variant is the most popular, as it is used by the author of `folke/tokyonight.nvim` himself.
> > While `https://github.com/tokyo-night/tokyo-night-vscode-theme` (enkia) was the original, you'll notice that `https://github.com/folke/tokyonight.nvim` is more widely adopted, and often serves as a reference for other implementations.
> > ```
[...]

PR #835, opened 2024-08-17
theme: Add Tokyo Night variants

foster-hangdaan:
This PR adds two variants of the Tokyo Night theme: Light and Storm

Ref: #827 

# Screenshots

**Tokyo Night Light**

image
image

**Tokyo Night Storm**

image
image

-----
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [x] My changes are visual; I've included before and after screenshots.
- [ ] I am blindly checking these off.
- [x] Any relevant issues or PRs have been linked to.
- [ ] This a draft PR; I need more time to finish it.

### DIFF
--- a/themes/doom-tokyo-night-light-theme.el
+++ b/themes/doom-tokyo-night-light-theme.el
@@ -1,0 +1,293 @@
+;;; doom-tokyo-night-light-theme.el --- inspired by VSCode's Tokyo Night theme -*- lexical-binding: t; no-byte-compile: t; -*-
+;;
+;; Added: August 17, 2024
+;; Author: Foster Hangdaan <https://code.fosterhangdaan.com/foster>
+;; Maintainer:
+;; Source: https://github.com/enkia/tokyo-night-vscode-theme
+;;
+;;; Commentary:
+;;; Code:
+
+(require 'doom-themes)
+
+
+;;
+;;; Variables
+
+(defgroup doom-tokyo-night-light-theme nil
+  "Options for doom-themes."
+  :group 'doom-themes)
+
+(defcustom doom-tokyo-night-light-brighter-modeline nil
+  "If non-nil, more vivid colors will be used to style the mode-line."
+  :group 'doom-tokyo-night-light-theme
+  :type 'boolean)
+
+(defcustom doom-tokyo-night-light-brighter-comments nil
+  "If non-nil, comments will be highlighted in more vivid colors."
+  :group 'doom-tokyo-night-light-theme
+  :type 'boolean)
+
+(defcustom doom-tokyo-night-light-comment-bg doom-tokyo-night-light-brighter-comments
+  "If non-nil, comments will have a subtle, darker background. Enhancing their legibility."
+  :group 'doom-tokyo-night-light-theme
+  :type 'boolean)
+
+(defcustom doom-tokyo-night-light-padded-modeline nil
+  "If non-nil, adds a 4px padding to the mode-line. Can be an integer to determine the exact padding."
+  :group 'doom-tokyo-night-light-theme
+  :type '(or integer boolean))
+
+
+;;
+;;; Theme definition
+
+(def-doom-theme doom-tokyo-night-light
+  "A light variant of Tokyo Night."
+
+  ; Color Scheme
+  ; gui       256
+  ; "#8c4351" "#8c4351" => This keyword, HTML elements, Regex group symbol, CSS units, Terminal Red
+  ; "#965027" "#965027" => Number and Boolean constants, Language support constants
+  ; "#8f5e15" "#8f5e15" => Function parameters, Regex character sets, Terminal Yellow
+  ; "#634f30" "#634f30" => Parameters inside functions (semantic highlighting only)
+  ; "#385f0d" "#385f0d" => Strings, CSS class names
+  ; "#33635c" "#33635c" => Object literal keys, Markdown links, Regex literal strings, Terminal Green
+  ; "#006c86" "#006c86" => Language support functions, CSS HTML elements
+  ; "#0f4b6e" "#0f4b6e" => Object properties, Regex quantifiers and flags, Markdown headings, Terminal Cyan, Markdown code, Import/export keywords
+  ; "#2959aa" "#2959aa" => Function names, CSS property names, Markdown Headings, Terminal Blue
+  ; "#5a3e8e" "#5a3e8e" => Control Keywords, Storage Types, Regex symbols and operators, HTML Attributes, Terminal Magenta
+  ; "#707280" "#707280" => Terminal White
+  ; "#40434f" "#40434f" => Markdown Text, HTML Text
+  ; "#343b58" "#343b58" => Editor foreground, Variables, Class names, Terminal black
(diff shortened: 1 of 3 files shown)


i-am-linja:
Hey @foster-hangdaan, I hate to nitpick but I believe the NVim port has italic comments and keywords across all variants. I can't submit a PR to the PR so could you please slant them? I _really_ like them.

foster-hangdaan:
> Hey @foster-hangdaan, I hate to nitpick but I believe the NVim port has italic comments and keywords across all variants. I can't submit a PR to the PR so could you please slant them? I _really_ like them.

The keywords and comments for the variants have been italicized. Note that I will not be modifying the original Night theme since it will be out of scope for this PR which already brings three new variants.

azzamsa:
But it looks very different than the upstream.

image


image

i-am-linja:
> But it looks very different than the upstream.

Worth noting: "upstream" is considered to be [the VSCode theme](https://github.com/tokyo-night/tokyo-night-vscode-theme), not the NVim port. Also, are you certain you're using the same variant in each app? Looks like Neovide is in Storm, whereas DOOM is in Night or Moon.

foster-hangdaan:
> But it looks very different than the upstream.
> 
> image
> 
> image

Thanks for the feedback. More screenshots of the Moon variant will be handy since I based the variant from the single screenshot that was provided on the Neovim port. I also didn't feel like crawling through the Lua code to obtain all the syntax highlighting rules.

Some things that *probably* won't be fixed base on those screenshots:

- Language-specific syntax highlighting. Take the Doom port, for example, Elisp's `use-package` and Rust's `let` are keywords and are simply following the global rule which specifies that keywords should be `magenta`. On the other hand we have the Neovim port, where Elisp's `use-package` is `magenta` (OK) *but* then uses a language-specific rule for Rust's `let` keyword which makes it `cyan`.

  I believe applying language-specific rules will be a community effort to pull off since it will require testing and modifying font faces for each major mode.

Some things that will be fixed:

- Strings are a slightly lighter tint of green.
- Types should be in blue and not grey. i.e. `IpAddr`.

azzamsa:
> but then uses a language-specific rule for Rust's let keyword which makes it cyan.

Yes, you are correct, those colors will be changed If I disable the lsp.

> Types should be in blue and not grey. i.e. IpAddr.

What about `base_url` and `http.port`? is it anguage-specific syntax highlighting.?

foster-hangdaan:
> > but then uses a language-specific rule for Rust's let keyword which makes it cyan.
> 
> Yes, you are correct, those colors will be changed If I disable the lsp.
> 
> > Types should be in blue and not grey. i.e. IpAddr.
> 
> What about `base_url` and `http.port`? is it anguage-specific syntax highlighting.?

I cannot check the Rust major mode since I don't have it nor its LSP installed but it's probably the same reason why the syntax highlighting for TypeScript methods and properties are also not being applied. On the screenshot below, `crypto.subtle` and `data.buffer` properties should be in `dark-cyan`:

image

When I inspect their text properties, they have no associated font-face:

image

Without a font face, there is nothing to target for highlighting, and it seems to fallback to the theme's foreground color.

So now you're probably asking why they don't have an associated font face? Not sure; that is beyond the limits of my Emacs-fu :cry:. My blame goes to the major mode and it is somehow not able identify the syntax. I've looked around and this problem to exists in several major modes.

TL;DR: It's an underlying Emacs issue (I think) and is out of the scope of a theme to fix.

foster-hangdaan:
Also @azzamsa and other Moon variant users, I would like to ask if it's really worth trying to copy the syntax highlighting rules of the Neovim port? Not only will [it will take a lot of work](https://github.com/doomemacs/themes/pull/835#issuecomment-2301779255), but I also think that the Doom highlighting rules are superior since they are more consistent across all major modes, whereas the Neovim port's rules are more arbitrary. Though, I do admit that the Neovim port is more colorful.

azzamsa:
> TL;DR: It's an underlying Emacs issue (I think) and is out of the scope of a theme to fix.

Ah, Okay I see.

> Though, I do admit that the Neovim port is more colorful.

It would be nice to look similar to neovim without enabling the lsp (language-specific rule).
We can safely ignore the color from language-specific rules.

prashantvithani:
A random suggestion for `storm` theme: Have `type` as `teal` instead of `base8`.

prashantvithani:
> A random suggestion for `storm` theme: Have `type` as `teal` instead of `base8`.

This is to differentiate `type` from `variable` as both are set to `base8` as of now.

foster-hangdaan:
> > A random suggestion for `storm` theme: Have `type` as `teal` instead of `base8`.
> 
> This is to differentiate `type` from `variable` as both are set to `base8` as of now.

Sounds good to me. I will also do the same for the light theme and change `type` to something more colorful than the current dull, grayish color.

themes/doom-tokyo-night-light-theme.el
```diff
@@ -0,0 +1,303 @@
...
+   (treemacs-git-modified-face :foreground green)
+   (treemacs-git-renamed-face :foreground yellow)
+
+   ;;; Magit
+   (magit-section-heading :foreground blue)
+   (magit-branch-remote   :foreground orange)
+   (magit-diff-our :foreground (doom-darken red 0.2) :background (doom-darken red 0.7))
+   (magit-diff-our-highlight :foreground red :background (doom-darken red 0.5) :weight 'bold)
+   (magit-diff-removed :foreground (doom-darken red 0.2) :background (doom-darken red 0.7))
+   (magit-diff-removed-highlight :foreground red :background (doom-darken red 0.5) :weight 'bold)
```
prashantvithani, line 237:
```suggestion
   (magit-diff-removed :foreground (doom-darken red 0.2) :background (doom-lighten red 0.7))
   (magit-diff-removed-highlight :foreground red :background (doom-lighten red 0.5) :weight 'bold)
```

**Before**:

No Highlight: image
Highlight: 
image

**After**:

No Highlight: 
image

Highlight: 
image

azzamsa:
What is the next steps to get this PR merged?

foster-hangdaan:
> What are the next steps to get this PR merged?

Unless there are further comments or change requests, then I'm done with this PR. It's all in the hands of @hlissner

azzamsa:
> > What are the next steps to get this PR merged?
> 
> Unless there are further comments or change requests, then I'm done with this PR. It's all in the hands of @hlissner

Thanks for all the efforts, I can use your repo. I think it is better to use a separate branch next time, so you have a clean `master` that matches the upstream.

```
(use-package doom-themes
  :ensure (doom-themes :repo "https://github.com/foster-hangdaan/doom-emacs-themes/")
  :config
  (load-theme 'doom-tokyo-night-moon t)
```

vincowl:
> > What are the next steps to get this PR merged?
> 
> Unless there are further comments or change requests, then I'm done with this PR. It's all in the hands of @hlissner

If everything is fine, could this PR be merged please ?

PR #874, opened 2025-10-20
Fix several malformed modeline colors. and diff-indicator-removed was vc-added instead of vc-deleted

wickedjargon:
This is a contribution made by an IRC user


hlissner:
I've merged this into 5890382 and ad9b1bd to conform their commit messages to our conventions. Thanks for bringing it to my attention!

